PR #10, opened 2016-02-06
Authorize included resources

### LINKED ISSUE #7: Authorization of included resources
Hey, I went through https://github.com/cerebris/jsonapi-resources/issues/16 and found your gem, looks nice. 
I have one question though, does the `operations_processor` approach take care of authorizing the included resources? Let me give you an example:

``` ruby
class PostResource < JSONAPI::Resource
  # read authorized for current_user
  has_many :comments 
end

class CommentResource < JSONAPI::Resource
  # read authorized for current_user
  has_one :user 
end

class UserResource < JSONAPI::Resource
  # read authorized ONLY when user == current_user
  has_many :comments 
end
```

If you try to get a post with the following request: `GET /posts/1?include=comments.user` we need to ensure that all comments and users relationships are authorized as well and not only the Post 1.

I think that @barelyknown's [approach](https://github.com/cerebris/jsonapi-resources/issues/16#issuecomment-108175995) take care of that by overwriting the [`JSONAPI::Resource#records_for`](https://github.com/cerebris/jsonapi-resources/blob/f0b951252e2efa4387b8676ce94aafd7f8337b8a/lib/jsonapi/resource.rb#L113-L117) method but I'm not sure with your approach.

What do you think? Thanks for your gem!

valscion:
Closes #7 

Logics to implement:

* [x] One-level deep `has_many` relationship (`?include=other-resources`) logic
* [x] One-level deep `has_one` relationship (`?include=another-resource`) logic
* [x] Multiple one-level relationships (`?include=other-resources,another-resource`) logic
* [x] Parse include directive graphs and authorize all relationships in the way (`?include=author.comments`)
* [x] Requests that have failed before authorization (e.g. due to validation errors) must not try to authorize include directives, too.
* [x] Add the source relation alongside the authorization checks

Attach those logics to all the end points which return records:

* [x] `GET /resources`
* [x] `GET /resources/1`
* [x] `PATCH /resources/1`
*  [x] `POST /resources`
* [x] `GET /resources/other-resources`
* [x] `GET /resources/another-resource`

Review notes:

* [x] Improve variable names in the operations processor recursive authorize method
* [x] Swap the authorizer method's argument order for consistency with other methods
* [ ] Don't use `public_send` on the model level as it bypasses restrictions imposed on the resource level
* [x] Simplify specs to just expect receiving `an_instance_of(Article)` style logic everywhere. It's not worth it to test for exact records, as we aren't able to do that on POST requests anyway.

### DIFF
--- a/lib/jsonapi/authorization/authorizing_operations_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_operations_processor.rb
@@ -30,11 +30,14 @@ module JSONAPI
 
       def authorize_include_directive
         return if @result.is_a?(::JSONAPI::ErrorsOperationResult)
-        resources = if @result.respond_to?(:resources)
-                      @result.resources
-                    else
-                      [@result.resource]
-                    end
+        resources = Array.wrap(
+          if @result.respond_to?(:resources)
+            @result.resources
+          elsif @result.respond_to?(:resource)
+            @result.resource
+          end
+        )
+
         resources.each do |resource|
           authorize_model_includes(resource._model)
         end
--- a/lib/jsonapi/authorization/authorizing_operations_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_operations_processor.rb
@@ -260,6 +260,7 @@ module JSONAPI
       def authorize_include_item(resource_klass, source_record, include_item)
         case include_item
         when Hash
+          # e.g. {articles: [:comments, :author]} when ?include=articles.comments,articles.author
           include_item.each do |rel_name, deep|
             authorize_include_item(resource_klass, source_record, rel_name)
             relationship = resource_klass._relationship(rel_name)
--- a/lib/jsonapi/authorization/authorizing_operations_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_operations_processor.rb
@@ -284,7 +284,7 @@ module JSONAPI
             related_record = source_record.public_send(
               relationship.relation_name(@operation.options[:context])
             )
-            authorizer.include_has_one_resource(related_record)
+            authorizer.include_has_one_resource(related_record) unless related_record.nil?
           when JSONAPI::Relationship::ToMany
             authorizer.include_has_many_resource(relationship.resource_klass._model_class)
           else
--- a/lib/jsonapi/authorization/authorizing_operations_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_operations_processor.rb
@@ -17,22 +17,30 @@ module JSONAPI
       set_callback :remove_to_many_relationship_operation, :before, :authorize_remove_to_many_relationship
       set_callback :remove_to_one_relationship_operation, :before, :authorize_remove_to_one_relationship
 
-      def authorize_find
-        authorizer.find(@operation.resource_klass._model_class)
+      [
+        :find_operation,
+        :show_operation,
+        :show_related_resource_operation,
+        :show_related_resources_operation,
+        :create_resource_operation,
+        :replace_fields_operation
+      ].each do |op_name|
+        set_callback op_name, :after, :authorize_include_directive
+      end
 
-        # Copy-pasted from jsonapi-resources.
-        # There really should be a better way.
-        resource_records = @operation.resource_klass.find(
-          @operation.resource_klass.verify_filters(@operation.filters, @operation.options[:context]),
-          context: @operation.options[:context],
-          include_directives: @operation.include_directives,
-          sort_criteria: @operation.sort_criteria,
-          paginator: @operation.paginator
-        )
-
-        resource_records.each do |resource|
+      def authorize_include_directive
+        resources = if @result.respond_to?(:resources)
+                      @result.resources
+                    else
+                      [@result.resource]
+                    end
+        resources.each do |resource|
           authorize_model_includes(resource._model)
         end
+      end
+
+      def authorize_find
+        authorizer.find(@operation.resource_klass._model_class)
       end
 
       def authorize_show
@@ -42,7 +50,6 @@ module JSONAPI
         )._model
 
         authorizer.show(record)
-        authorize_model_includes(record)
       end
 
       def authorize_show_relationship
@@ -79,7 +86,6 @@ module JSONAPI
         source_record = source_resource._model
         related_record = related_resource._model unless related_resource.nil?
         authorizer.show_related_resource(source_record, related_record)
-        authorize_model_includes(source_record)
       end
 
       def authorize_show_related_resources
@@ -89,7 +95,6 @@ module JSONAPI
         )._model
 
         authorizer.show_related_resources(source_record)
-        authorize_model_includes(source_record)
       end
 
       def authorize_replace_fields
@@ -99,7 +104,6 @@ module JSONAPI
         )._model
 
         authorizer.replace_fields(source_record, related_models)
-        authorize_model_includes(source_record)
       end
 
       def authorize_create_resource
--- a/lib/jsonapi/authorization/authorizing_operations_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_operations_processor.rb
@@ -264,17 +264,17 @@ module JSONAPI
           include_item.each do |rel_name, deep|
             authorize_include_item(resource_klass, source_record, rel_name)
             relationship = resource_klass._relationship(rel_name)
-            new_resource_klass = relationship.resource_klass
+            next_resource_klass = relationship.resource_klass
             Array.wrap(
               source_record.public_send(
                 relationship.relation_name(@operation.options[:context])
               )
-            ).each do |new_source_record|
-              deep.each do |new_include_item|
+            ).each do |next_source_record|
+              deep.each do |next_include_item|
                 authorize_include_item(
-                  new_resource_klass,
-                  new_source_record,
-                  new_include_item
+                  next_resource_klass,
+                  next_source_record,
+                  next_include_item
                 )
               end
             end
--- a/lib/jsonapi/authorization/authorizing_operations_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_operations_processor.rb
@@ -97,6 +97,7 @@ module JSONAPI
         )._model
 
         authorizer.replace_fields(source_record, related_models)
+        authorize_model_includes(source_record)
       end
 
       def authorize_create_resource
@@ -242,8 +243,8 @@ module JSONAPI
       end
 
       def authorize_model_includes(source_record)
-        if @operation.include_directives
-          @operation.include_directives.model_includes.each do |include_item|
+        if @request.include_directives
+          @request.include_directives.model_includes.each do |include_item|
             authorize_include_item(@operation.resource_klass, source_record, include_item)
           end
         end
--- a/lib/jsonapi/authorization/authorizing_operations_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_operations_processor.rb
@@ -19,7 +19,20 @@ module JSONAPI
 
       def authorize_find
         authorizer.find(@operation.resource_klass._model_class)
-        authorize_model_includes
+
+        # Copy-pasted from jsonapi-resources.
+        # There really should be a better way.
+        resource_records = @operation.resource_klass.find(
+          @operation.resource_klass.verify_filters(@operation.filters, @operation.options[:context]),
+          context: @operation.options[:context],
+          include_directives: @operation.include_directives,
+          sort_criteria: @operation.sort_criteria,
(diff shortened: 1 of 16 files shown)


thibaudgg:
> ~~POST /resources~~ This seems to be impossible. It can be done later if we figure it out.

Why would it be impossible? We actually have that case where we want to create an object and get related resources in the response. Something like: `POST /articles?include=author.articles` (create an article and get all existing articles from the same author). Why would it be an issue? We have the `author` from relationship data of the `POST` request and we can check `index?` on the `Article` class easily.

Thanks for working on this!

valscion:
> Why would it be impossible?

The record itself isn't created by by the time the operation processor runs for `create_action`, so it gets quite difficult to try to get the related models underneath before the root object is built.

The policy scoping will be applied still for the `has_many` resources, but this PR doesn't yet support stopping the entire POST request if the user wouldn't be authorized to include the `author.articles` records.

We can always tackle that issue on a follow-up PR. This risks getting too large and difficult to review.

valscion:
> and we can check index? on the Article class easily.

Yes, but we'd first have to go through the author and use `show?` on it. Currently this functionality relies on the record being present by the time we go and fetch all the relationships etc. There might be a possibility to still get an `index?` authorization without having the record available yet, but `show?` for `has_one` associations is a no-go.

The way the include directives are parsed is by going up the hash tree available to operations processor via `@request.include_directives`. It relies on going through all the steps. And it might actually even crash for nils currently if a `has_one` association is null in a graph scenario :sweat_smile: `author.comments` will crash if `author` is nil.

thibaudgg:
I think what we discussed in #7 was to authorize all included resource from inside the `records_for` method and not at the operation level like your current implementation here. Handling `POST` request would not be an issue if all the logic is handled at the JSONAPI::Resource#records_for`. Don't you think?

valscion:
> I think what we discussed in #7 was to authorize all included resource from inside the records_for method

Yeah you're right. I wanted to see what kinds of requests trigger that `records_for` method. It seems that [basic relationship operations also trigger it](https://travis-ci.org/venuu/jsonapi-authorization/jobs/107558599#L601-L618), and we can't really get enough context on the resource level `#records_for` on what kind of request is incoming to call only the correct authorizer function.

I think that my approach here, which relies on `before` hooks is a bit flawed. As inclusion of related resources happens *after* the operation is complete, maybe I should also just authorize the inclusion in  `after` callbacks. This way I might even have the resource available somehow in the operation level instance variables, and it would work for POST requests, too.

I'd like to keep the authorizing logic inside the operations processor if at all possible, and if we do get proper insight to requests on resource level callbacks and methods, we might want to look into moving every authorization inside it.

thibaudgg:
Ok it makes sense to keep all the authorization logic inside the operations processor, in that case the `after` callback could be the way to go. :+1:

valscion:
Ok, now it's done but the spec structure (and the code, too, at least a bit) leaves a lot to be desired. We might also want to look at how to test the related resource includes, too, without having to rely on ugly self-referencing association hacks.

valscion:
And oh yeah, this code will most likely crash if:

* A relationship is `nil` before the leaf node of include directives: `?include=author.comments` and `author` relationship is not present.
* The result has errored before we reach include directive check, e.g. when the original request was unauthorized already

Those cases should be tackled, but I haven't yet had the time to cover them in tests.

valscion:
@thibaudgg, what kind of authorization checks do you think should happen in this case?

```
GET /articles/1?include=author.comments
```

and `Article.find(1).author == nil`?

Currently I have a working implementation where the authorization short circuits immediately for `nil` has_one relation. This would mean that the authorizations aren't called at all, as the returned JSON won't have the included resources nevertheless. That is how I vision it would work for this case, too:

```
GET /articles/1?include=comments.author
```

and `Article.find(1).comments.size == 0`

valscion:
> The result has errored before we reach include directive check, e.g. when the original request was unauthorized already

Actually, we'll never reach that case as if any of the `before` callbacks have failed, `after` callbacks aren't ran at all AFAIK.

> A relationship is nil before the leaf node of include directives: `?include=author.comments` and author relationship is not present.

This case is now fixed, too. It short-circuits the authorization logic on the `author` level already.

thibaudgg:
Looks good to me, I'll give it a try on our app (still need to define my Policies first). :+1:

lib/jsonapi/authorization/default_pundit_authorizer.rb
```diff
@@ -190,6 +190,36 @@ def remove_to_many_relationship(source_record, related_record)
...
+      # directive has been travelled completely.
+      #
+      # We can't pass all the records of a +has_many+ association here due to
+      # performance reasons, so the class is passed instead.
+      #
+      # ==== Parameters
+      #
+      # * +record_class+ - The underlying record class for the relationships
+      #                    resource.
+      def include_has_many_resource(record_class)
```
valscion: I might be able to pass the source relation record in these `include` methods here, too. Should I?

thibaudgg: The idea is to be able to apply Pundit::Scope here, right?

valscion: Pundi::Scope is applied already in the resource level, so we don't have to do it here. Passing the source relation record here would merely allow custom logic for the authorization check, e.g. it would allow a case such as

```ruby
if source_relation_record.is_a?(Article)
  ::Pundit.authorize(user, record_class, 'include_with_article?')
else
  ::Pundit.authorize(user, record_class, 'index?')
end
```

Right now this authorization method does not have context of what was the source of the include query — that is, if we got here through `?include=author.comments` (`record_class` would be `Comment` in that case), this method does not know that we're including comments for an author record.

lib/jsonapi/authorization/authorizing_operations_processor.rb
```diff
@@ -17,6 +17,28 @@ class AuthorizingOperationsProcessor < ::ActiveRecordOperationsProcessor
...
+        :replace_fields_operation
+      ].each do |op_name|
+        set_callback op_name, :after, :authorize_include_directive
+      end
+
+      def authorize_include_directive
+        resources = if @result.respond_to?(:resources)
+                      @result.resources
+                    else
+                      [@result.resource]
```
thibaudgg: If the request result is a `JSONAPI::ErrorsOperationResult` (e.g. `POST` request with invalid fields) we get `undefined method 'resource' for #<JSONAPI::ErrorsOperationResult>` so it should be something like that:

``` ruby
resources = Array.wrap(
  if @result.respond_to?(:resources)
    @result.resources
  elsif @result.respond_to?(:resource)
    @result.resource
  end
)
```

lib/jsonapi/authorization/default_pundit_authorizer.rb
thibaudgg: Ok makes sense to add it then. :+1:

lib/jsonapi/authorization/authorizing_operations_processor.rb
valscion: Okay awesome! This was exactly the case I figured we'd hit. I'll try to come up with a test for it, but I'll fix it in the meantime so you can keep testing this branch :)

thibaudgg: Perfect, thanks!

thibaudgg:
Just hit the [`# TODO: find_by_key?` comment](https://github.com/venuu/jsonapi-authorization/blob/authorize-included-resources/lib/jsonapi/authorization/authorizing_operations_processor.rb#L246-L247) as some of our Resources use `external_id` as primary key. I've solved it with:

``` ruby      
      def resource_class_for_relationship(assoc_name)
        @operation.resource_klass._relationship(assoc_name).resource_klass
      end

      def related_models
        data = @operation.options[:data]
        return [] if data.nil?

          data[rel_type].flat_map do |assoc_name, assoc_ids|
            resource_class = resource_class_for_relationship(assoc_name)
            resource_class.find_by_key(assoc_ids, context: @operation.options[:context])
          end
        end
      end
```

Should I create a dedicated issue/PR for it?

valscion:
A PR would be great! If you can tweak the dummy app in the tests to match the `external_id` primary key used in your resources, it'd be even better :+1:

valscion:
Ok the functionality seems to be complete now, alongside with passing the relationship source record to the authorizer. Phew. The specs are a bit messy right now, but at least they should be complete.

lib/jsonapi/authorization/authorizing_operations_processor.rb
```diff
@@ -225,6 +248,57 @@ def related_models
...
+        case include_item
+        when Hash
+          include_item.each do |rel_name, deep|
+            authorize_include_item(resource_klass, source_record, rel_name)
+            relationship = resource_klass._relationship(rel_name)
+            new_resource_klass = relationship.resource_klass
+            Array.wrap(
+              source_record.public_send(
+                relationship.relation_name(@operation.options[:context])
+              )
```
valscion: This `public_send` straight to the record will short-circuit policy scoping on the resource level. And that's bad.

lib/jsonapi/authorization/default_pundit_authorizer.rb
```diff
@@ -190,6 +190,40 @@ def remove_to_many_relationship(source_record, related_record)
...
+      # directive has been travelled completely.
+      #
+      # ==== Parameters
+      #
+      # * +related_record+ - The associated record to return
+      # * +source_record+ — The source relationship record, e.g. an Article in
+      #                     article.author check
+      def include_has_one_resource(related_record, source_record)
+        ::Pundit.authorize(user, related_record, 'show?')
+      end
```
lime: I'd swap the order of the arguments here to be more consistent with the rest of the methods.

lib/jsonapi/authorization/default_pundit_authorizer.rb
```diff
@@ -190,6 +190,40 @@ def remove_to_many_relationship(source_record, related_record)
...
+      # This will be called for each has_one relationship if the include goes
+      # deeper than one level until some authorization fails or the include
+      # directive has been travelled completely.
+      #
+      # ==== Parameters
+      #
+      # * +related_record+ - The associated record to return
+      # * +source_record+ — The source relationship record, e.g. an Article in
+      #                     article.author check
+      def include_has_one_resource(related_record, source_record)
```
lime: I'd swap the order of the arguments here to be more consistent with the rest of the methods.

lib/jsonapi/authorization/authorizing_operations_processor.rb
thibaudgg: @valscion this had not be added in that branch and it's causing issue on our spec still. Any reason for it?

``` ruby
resources = Array.wrap(
  if @result.respond_to?(:resources)
    @result.resources
  elsif @result.respond_to?(:resource)
    @result.resource
  end
)
``` 
still fixes it.

thibaudgg:
I have tested this branch on merged with #11 and it works fine with [that fix](https://github.com/electric-feel/jsonapi-authorization/commit/bdf36b7e78c6e664c920a5bd7694f4f1ccec1aed), all our 336 API specs are green :green_heart:.

lime:
Great stuff, both of you! :+1: 

There is most definitely still work that could be done here, Currently `authorize_include_item` is such a beast that it's bound to come back and bite us. :grimacing: Still, it might be best to get this merged and keep coding on top of it, so as to lessen the risk of conflicts for other PRs.

valscion:
Fixed the merge conflicts here after #11 was merged :)

thibaudgg:
Just tested that branch and everything still green for us. I'll use that one until it's merged. 
Thanks a lot for your great work here. :tada:

valscion:
We discussed this with @lime today. Sorry about the delay, I'll merge this now and release a new version soon after — I'll put up an issue about the `public_send` with relationships. I could need a little help with that, and it'll be much easier to do when the amount of code that needs to be changed is much smaller than this PR.

valscion:
v0.6.0 pushed out :) feel free to use that one @thibaudgg, it should contain all the necessary bits :)

valscion:
> I'll put up an issue about the public_send with relationships. I could need a little help with that, and it'll be much easier to do when the amount of code that needs to be changed is much smaller than this PR.

Actually, I think that currently we are doing the right thing when fetching the relationships directly from the record instead of going through resource classes. If we hit a case for trying to sideload a record we're not allowed to see with include directive, we should authorize for them even if we'd get a 404 when we'd try to show that resource via a normal show operation.

thibaudgg:
:+1: using 0.6 now! Thanks!

PR #11, opened 2016-02-08
Add support for custom primary key on related_models

thibaudgg:
Discussed on https://github.com/venuu/jsonapi-authorization/pull/10#issuecomment-181308985

(Edit by @valscion: Closes #13)

### DIFF
--- a/spec/requests/resource_operations_spec.rb
+++ b/spec/requests/resource_operations_spec.rb
@@ -34,13 +34,13 @@ describe 'Resource operations', type: :request do
 
       it 'returns results limited by policy scope' do
         expect(json_data.length).to eq(1)
-        expect(json_data.first["id"]).to eq(article.id.to_s)
+        expect(json_data.first["id"]).to eq(article.external_id)
       end
     end
   end
 
   describe 'GET /articles/:id' do
-    subject(:last_response) { get("/articles/#{article.id}") }
+    subject(:last_response) { get("/articles/#{article.external_id}") }
     let(:policy_scope) { Article.all }
 
     context 'unauthorized for show' do
@@ -70,7 +70,7 @@ describe 'Resource operations', type: :request do
   end
 
   describe 'POST /articles' do
-    subject(:last_response) { post("/articles", '{ "data": { "type": "articles" } }') }
+    subject(:last_response) { post("/articles", '{ "data": { "id": "1", "type": "articles" } }') }
 
     context 'unauthorized for create_resource' do
       before { disallow_operation('create_resource', Article, []) }
@@ -88,14 +88,14 @@ describe 'Resource operations', type: :request do
       <<-EOS.strip_heredoc
       {
         "data": {
-          "id": "#{article.id}",
+          "id": "#{article.external_id}",
           "type": "articles"
         }
       }
       EOS
     end
 
-    subject(:last_response) { patch("/articles/#{article.id}", json) }
+    subject(:last_response) { patch("/articles/#{article.external_id}", json) }
     let(:policy_scope) { Article.all }
 
     context 'authorized for replace_fields' do
@@ -120,7 +120,7 @@ describe 'Resource operations', type: :request do
   end
 
   describe 'DELETE /articles/:id' do
-    subject(:last_response) { delete("/articles/#{article.id}") }
+    subject(:last_response) { delete("/articles/#{article.external_id}") }
     let(:policy_scope) { Article.all }
 
     context 'unauthorized for remove_resource' do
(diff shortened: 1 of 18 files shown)


valscion:
Weird. The `tricky_operations_spec.rb` seems to fail when it is run on its own with `bundle exec rspec spec/requests/tricky_operations_spec.rb`.

Also, now I remember why we used `find`. `find_by_key` returns *only the first element* of the relationship — so when we try to specify it for multiple resources, it doesn't call the authorizer with them all but only with the first one.

I created a PR where I implemented the tricky operations specs, #13. The diff is interesting, as it has the checks done for the authorizer contract instead of Pundit. We actually forgot to update the tricky operations specs to use that strategy and I'm sorry for that :confused: 

If you modify your PR to use authorization level stubs, it will fail, as [this check](https://github.com/venuu/jsonapi-authorization/pull/13/files#diff-047edb8d8b76bd3cdd1fa67665792a1bR88) in the #13 PR wouldn't be called with all the new comments, but only the first one:

```
before { allow_operation('replace_fields', article, new_comments) }
```

It might be an issue in JR gem itself — if there would be a way around it, it'd be great.

I'll check again on this PR tomorrow.

thibaudgg:
Ok, I'll have a look again tomorrow as well.

thibaudgg:
Hei @valscion, I rebased my PR on your `tricky-specs` branch and updated the `related_models` query to make all specs pass again. It should be good now.

spec/requests/resource_operations_spec.rb
```diff
@@ -70,7 +70,7 @@
   end
 
   describe 'POST /articles' do
-    subject(:last_response) { post("/articles", '{ "data": { "type": "articles" } }') }
+    subject(:last_response) { post("/articles", '{ "data": { "id": "1", "type": "articles" } }') }
```
valscion: I don't quite follow why you must specify an `id` here? Is it that this PR changes the `Article` resource to have its ID generated client-side like as an UUID? Auto-increment-like behavior no longer works, I assume.

thibaudgg: The reason is as we're forcing the presence of `external_id` in the db we need to set it from somewhere. In a production env it could be set from a DB pgsql function or something (UUID). 

> Auto-increment-like behavior no longer works, I assume.

You're correct, but I don't think it's a big issue for the specs right?

valscion: Yeah I'm fine with this :+1: It didn't come to my mind that it could be set on the database side, too. I agree it's not necessary to complicate the test database setup with ID generation, so it's ok that the ID is specified here.

Should you, though, use another id as `1` here? Doesn't the fixtures already specify an article with the similar ID? I wonder why this request didn't fail if there already was an article in the DB :confused:

valscion:
Super amazing work, thank you! :heart_eyes:

valscion:
Super amazing work, thank you! :heart_eyes: 

Besides the one oddity with the ID in the POST spec, I have no issues whatsoever with this PR. Amazing to have you try this gem on your real, large API :+1:

thibaudgg:
@valscion thanks my pleasure, I'll update the ID value with something more specific.

I just get an issue with a polymorphic association, so maybe hold a little bit before merging. I'm not sure that it's directly link to that change or if polymorphism will need a separated PR.

valscion:
> I'm not sure that it's directly link to that change or if polymorphism will need a separated PR.

I have a feeling that we'll get issues with polymorphic associations elsewhere, too, which have been present even before this PR. But I'll wait :)

thibaudgg:
Ok, I fixed my polymorphic issue by updating the `related_models` method with:

``` ruby
def resource_class_for_relationship(assoc_name)
  @operation.resource_klass.resource_for(assoc_name.to_s)
end

def related_models
  data = @operation.options[:data]
  return [] if data.nil?

    data[rel_type].flat_map do |assoc_name, assoc_value|
      case assoc_value
      when Hash # Polymorphic relationship
        resource_class = resource_class_for_relationship(assoc_value[:type])
        resource_class.find_by_key(assoc_value[:id], context: @operation.options[:context])._model
      else
        resource_class = resource_class_for_relationship(assoc_name)
        primary_key = resource_class._primary_key
        resource_class._model_class.where(primary_key => assoc_value)
      end
    end
  end
end
```

I think this can be done in this PR, do you already have something in mind for a polymorphic association in the dummy app? A `Tag` object for comments and articles?

valscion:
> I think this can be done in this PR, do you already have something in mind for a polymorphic association in the dummy app? A Tag object for comments and articles?

Feel free to do it here :+1: A `Tag` object seems reasonable.

thibaudgg:
Polymorphic relationship support added, tested with all resources in our app, no more spec failures.

valscion:
This is _so_ good, thank you! Everything seems to be in order here.

thibaudgg:
Awesome, thanks!

Issue #17, opened 2016-04-19
Relax jsonapi-resources version constraint?


arcreative:
Can you drop the `.0` off of jsonapi-resources `0.7.0` please?  Trying to get this working with the new beta, but bundler is yelling about it!

valscion:
We intentionally wanted to constrain this gem to a single version as we're reaching to quite a lot of JR internals. That being said, I don't think the beta has changed the internals much..? @arcreative can you create a PR which would run the tests against the beta version of JR, and relax the gemspec constraints? If all the tests pass, I suppose we'd be confident enough to allow running beta, too.

arcreative:
I think that's reasonable--I don't really think the beta is going to fix my current problems, unfortunately...

lime:
FYI @arcreative, we've now released version 0.6.1 of the gem with a more permissive version constraint, you might want to take it for a spin. :)

arcreative:
@lime Awesome, thanks!  Probably come in handy with an impending Rails 5 upgrade as well :-)

PR #19, opened 2016-04-25
Relax jsonapi-resources version constraint

lime:
As requested in #17, this allows for using the 0.7.1.beta1 version of jsonapi-resources.

Not sure if we should use the permissive `"~> 0.7.0"` or limit it to only `">= 0.7.0", "<= 0.7.1.beta1"`.

On one hand, we depend on so much internal API that it's not unthinkable for even a patch version to contain breaking changes for us. On the other hand, it's frustrating for users to have to wait for our version bump every time JR is updated.

I'd go with the more permissive constraint, but try to list each supported version of JR in our build matrix. So if you notice a new version is released and not yet included in the build matrix, please send a PR. :)


Gemfile
```diff
@@ -4,14 +4,24 @@ gemspec
...
+  gem 'rails', "~> #{rails_version}"
+end
+
+case jsonapi_resources_version
+when 'master'
+  gem 'jsonapi-resources', git: 'https://github.com/cerebris/jsonapi-resources.git'
+when 'default'
+  gem 'jsonapi-resources', '0.7.0'
+else
+  gem 'jsonapi-resources', "~> #{jsonapi_resources_version}"
```
valscion: Should this be a fixed number to ensure we test against the correct specified version?

lime: Good point. I guess it makes sense to have a `~>` for Rails, but for JR the whole point was to test the exact version.

valscion: Yeah, `~>` version constraint works for Rails, but maybe it could also be the exact environment variable. That way we could just specify it with the `~>` constraint in build matrix so it'd be more explicit on what we're testing for.

valscion:
Looks good :+1: :boat:

PR #26, opened 2016-08-03
Updating config example to work for operations processor branch.

robomc:


valscion:
Thanks! I added this in #27 to point to the renamed authorizing processor class together with other README changes. If you feel there's something missing from the README, please tell :)

PR #32, opened 2016-08-22
Added ability to custom pundit_user

coryodaniel:
* Added tests
* Defaults to :user for backwards combat
* Can specify a symbol or a proc
* Updated README

```ruby
JSONAPI::Authorization.configure do |config|
  config.pundit_user = :current_user
  # or
  config.pundit_user = ->(context){ context[:current_user] }
  # or if you want the whole context available for pundit to use
  config.pundit_user = ->(context){ context }
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,16 @@ JSONAPI::Authorization.configure do |config|
 end
 ```
 
+By default JSONAPI::Authorization uses the `:user` key from the JSONAPI context hash as the Pundit user. If you would like to use `:current_user` or some other key, it can be configured as well.
+
+```ruby
+JSONAPI::Authorization.configure do |config|
+  config.pundit_user = :current_user
+  # or a block can be provided
+  config.pundit_user = ->(context){ context[:current_user] }
+end
+```
+
 ## Troubleshooting
 
 ### "Unable to find policy" exception for a request
--- a/spec/dummy/db/schema.rb
+++ b/spec/dummy/db/schema.rb
@@ -25,8 +25,8 @@ ActiveRecord::Schema.define(version: 20160125083537) do
   end
 
   create_table "tags", force: :cascade do |t|
-    t.integer "taggable_id"
     t.string  "taggable_type"
+    t.integer "taggable_id"
   end
 
   create_table "users", force: :cascade do |t|
--- a/lib/jsonapi/authorization/configuration.rb
+++ b/lib/jsonapi/authorization/configuration.rb
@@ -4,9 +4,19 @@ module JSONAPI
   module Authorization
     class Configuration
       attr_accessor :authorizer
+      attr_accessor :pundit_user
 
       def initialize
-        self.authorizer = ::JSONAPI::Authorization::DefaultPunditAuthorizer
+        self.authorizer  = ::JSONAPI::Authorization::DefaultPunditAuthorizer
+        self.pundit_user = :user
+      end
+
+      def user_context(context)
+        if pundit_user.is_a?(Symbol)
+          context[pundit_user]
+        else
+          pundit_user.call(context)
+        end
       end
     end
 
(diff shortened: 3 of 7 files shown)


valscion:
Hi, and thanks for the PR! This looks like the correct approach :+1: There seems to be some unrelated changes in your diff which caused some delay in responding, forgive me :). I'll add inline comments on the parts that I'm talking about.

.ruby-version
```diff
@@ -1 +1 @@
-2.1.2
+2.2.2
```
valscion: Could you refrain from bumping the `.ruby-version`? That change feels orthogonal to this PR and is in fact not required.

README.md
```diff
@@ -96,6 +96,9 @@ You can use a custom authorizer class by specifying a configure block in an init
 ```ruby
 JSONAPI::Authorization.configure do |config|
   config.authorizer = MyCustomAuthorizer
+  config.pundit_user = :user
+  # or
+  config.pundit_user = ->(context){ context[:current_user] }
```
valscion: These changes could be described a little better, as why they are necessary and that they aren't required even if you're using a custom authorizer. I can do that once this PR is merged as well :+1:

lib/jsonapi/authorization.rb
```diff
@@ -7,6 +7,14 @@
...
 module JSONAPI
   module Authorization
+    def user_context(context)
+      if JSONAPI::Authorization.configuration.pundit_user.is_a?(Symbol)
+        context[JSONAPI::Authorization.configuration.pundit_user]
+      else
+        JSONAPI::Authorization.configuration.pundit_user.call(context)
+      end
+    end
+    module_function :user_context
```
valscion: Do you think that this is the best place for this method? One option would be to move it to `Authorization::Configuration` class, so that calling it would look like this:

```rb
user_context = JSONAPI::Authorization.configuration.user_context(context)
```

It's a little bit more typing but it has that feeling it is a part of the configuration. What do you think?

spec/dummy/db/schema.rb
```diff
@@ -1,4 +1,3 @@
-# encoding: UTF-8
```
valscion: I suppose removal of this comment was because you used a different Ruby version? Could you remove the changes done to `schema.rb` from this PR as they're not really relevant here.

spec/jsonapi/authorization_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'spec_helper'
+
+RSpec.describe JSONAPI::Authorization::DefaultPunditAuthorizer do
```
valscion: I suppose these specs actually don't test `DefaultPunditAuthorizer` so this describe argument is misleading, don't you think? And if you decide to go the way of moving `user_context` method to `Configuration` class as commented in https://github.com/venuu/jsonapi-authorization/pull/32/files#r76201264, these tests could also be moved to `spec/jsonapi/authorization/configuration_spec.rb` file.

spec/jsonapi/authorization_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  after{
+    # Set this back to the default after each
+    JSONAPI::Authorization.configuration.pundit_user = :user
+  }
```
valscion: What do you feel about using `after do ... end` for multiline blocks?

```rb
after do
  # Set this back to the default after each
  JSONAPI::Authorization.configuration.pundit_user = :user
end
```

coryodaniel:
Sounds good. I left on vacay today. Will update when I get back.

valscion:
Have a great vacation! Don't stress too much about this :)

coryodaniel:
Hey @valscion LMK if the README reads any better! :D

valscion:
Looks great, thanks! I wonder why you commented out the `config.middleware.delete "ActionDispatch::BestStandardsSupport"` line? I don't know myself why that was there in the first place, it probably originated from scaffolding the dummy application in the first place.

Anyway, it doesn't stop this from being merged. Thank you for your contribution!

valscion:
Thanks again @coryodaniel, I released this in version 0.8.1 :rocket:

Issue #30, opened 2016-08-07
API proposal for authorizing relationships


valscion:
UPDATE on 2016-10-29:

The approach I want to take on authorizing relationships relies on always using a whitelist instead of blacklist. This means that:
- There won't be a generic `allow_relationship?` method that would handle all relationships.
- Using `update?` after trying to set a foreign key won't work, as it is a blacklist approach.

So even at the cost of some policies becoming huge, I want to keep each method as small and unambiguous as possible. That means methods named like this:

``` rb
class CommentPolicy
  def allow_relationship_article?(article)
    # ...
  end
end
```

I haven't yet made up my mind on which side of the association the authorization should happen, but I'd like that choice to be abstracted away from the [`AuthorizingProcessor`](https://github.com/venuu/jsonapi-authorization/blob/v0.8.1/lib/jsonapi/authorization/authorizing_processor.rb). It could live inside [`DefaultPunditAuthorizer`](https://github.com/venuu/jsonapi-authorization/blob/v0.8.1/lib/jsonapi/authorization/default_pundit_authorizer.rb), if it would make sense, or we could come up with a new place for that logic to live in.

See [my comment about this](https://github.com/venuu/jsonapi-authorization/issues/30#issuecomment-250911834) for more details, and to continue the discussion.

<details>
<summary>

Open this to read the old proposal that kicked off the conversation</summary>

For a create request that looks like this:

``` json
{
  "data": {
    "id": "post-1",
    "type": "posts",
    "relationships": {
      "author": {
        "data": {
          "id": "user-1", "type": "user"
        }
      },
      "comments": {
        "data": [
          { "id": "comment-1", "type": "comments" },
          { "id": "comment-2", "type": "comments" }
        ]
      },
      "tags": {
        "data": [
          { "id": "tag-1", "type": "tags" },
          { "id": "tag-2", "type": "tags" }
        ]
      }
    }
  }
}
```

We authorize with these methods:
1. `PostPolicy.new(post_1, current_user).create?`
2. `PostPolicy.new(post_1, current_user).associate_with_author(user_1)?`
3. `PostPolicy.new(post_1, current_user).associate_with_comments([comment_1, comment_2])?`
4. `PostPolicy.new(post_1, current_user).associate_with_tags([tag_1, tag_2])?`

For this request, no matter if it's an update or a create call, we are able to only use `PostPolicy` to authorize for every association.

[...]

PR #40, opened 2016-12-12
Authorize relationship actions with custom policy methods

matthias-g:
This PR is to discuss a proposal for how the authorization of relationships could be handled (see #30).

### DIFF
--- a/lib/jsonapi/authorization/default_pundit_authorizer.rb
+++ b/lib/jsonapi/authorization/default_pundit_authorizer.rb
@@ -148,14 +148,14 @@ module JSONAPI
       # * +new_related_records+ - The new records to be added to the association
       # * +relationship_type+ - The relationship type
       def create_to_many_relationship(source_record, new_related_records, relationship_type)
-        policy = Pundit.policy(user, source_record)
-        relationship_method = 'allow_relationship_' + relationship_type.to_s + '?'
-        if policy.respond_to?(relationship_method)
-          allowed = policy.send(relationship_method, new_related_records)
-        else
-          allowed = policy.update?
-        end
-        raise Pundit::NotAuthorizedError unless allowed
+        policy = ::Pundit.policy(user, source_record)
+        relationship_method = "allow_relationship_#{relationship_type}?"
+        allowed = if policy.respond_to?(relationship_method)
+                    policy.send(relationship_method, new_related_records)
+                  else
+                    policy.update?
+                  end
+        raise ::Pundit::NotAuthorizedError unless allowed
       end
 
       # <tt>PATCH /resources/:id/relationships/other-resources</tt>
--- a/lib/jsonapi/authorization/default_pundit_authorizer.rb
+++ b/lib/jsonapi/authorization/default_pundit_authorizer.rb
@@ -134,18 +134,8 @@ module JSONAPI
       # * +new_related_record+ - The new record replacing the old record
       # * +relationship_type+ - The relationship type
       def replace_to_one_relationship(source_record, new_related_record, relationship_type)
-        policy = ::Pundit.policy(user, source_record)
         relationship_method = "replace_#{relationship_type}?"
-        if policy.respond_to?(relationship_method)
-          unless policy.public_send(relationship_method, new_related_record)
-            raise ::Pundit::NotAuthorizedError,
-                  query: relationship_method,
-                  record: source_record,
-                  policy: policy
-          end
-        else
-          ::Pundit.authorize(user, source_record, 'update?')
-        end
+        authorize_relationship_operation(source_record, relationship_method, new_related_record)
       end
 
       # <tt>POST /resources/:id/relationships/other-resources</tt>
@@ -158,18 +148,8 @@ module JSONAPI
       # * +new_related_records+ - The new records to be added to the association
       # * +relationship_type+ - The relationship type
       def create_to_many_relationship(source_record, new_related_records, relationship_type)
-        policy = ::Pundit.policy(user, source_record)
         relationship_method = "add_to_#{relationship_type}?"
-        if policy.respond_to?(relationship_method)
-          unless policy.public_send(relationship_method, new_related_records)
-            raise ::Pundit::NotAuthorizedError,
-                  query: relationship_method,
-                  record: source_record,
-                  policy: policy
-          end
-        else
-          ::Pundit.authorize(user, source_record, 'update?')
-        end
+        authorize_relationship_operation(source_record, relationship_method, new_related_records)
       end
 
       # <tt>PATCH /resources/:id/relationships/other-resources</tt>
@@ -183,18 +163,8 @@ module JSONAPI
       #   association
       # * +relationship_type+ - The relationship type
       def replace_to_many_relationship(source_record, new_related_records, relationship_type)
-        policy = ::Pundit.policy(user, source_record)
         relationship_method = "replace_#{relationship_type}?"
-        if policy.respond_to?(relationship_method)
-          unless policy.public_send(relationship_method, new_related_records)
-            raise ::Pundit::NotAuthorizedError,
-                  query: relationship_method,
-                  record: source_record,
-                  policy: policy
-          end
-        else
-          ::Pundit.authorize(user, source_record, 'update?')
-        end
+        authorize_relationship_operation(source_record, relationship_method, new_related_records)
       end
 
       # <tt>DELETE /resources/:id/relationships/other-resources</tt>
@@ -209,18 +179,8 @@ module JSONAPI
       # * +related_record+ - The record which will be disassociated from +source_record+
       # * +relationship_type+ - The relationship type
       def remove_to_many_relationship(source_record, related_record, relationship_type)
-        policy = ::Pundit.policy(user, source_record)
         relationship_method = "remove_from_#{relationship_type}?"
-        if policy.respond_to?(relationship_method)
-          unless policy.public_send(relationship_method, related_record)
-            raise ::Pundit::NotAuthorizedError,
-                  query: relationship_method,
-                  record: source_record,
-                  policy: policy
-          end
-        else
-          ::Pundit.authorize(user, source_record, 'update?')
-        end
+        authorize_relationship_operation(source_record, relationship_method, related_record)
       end
 
       # <tt>DELETE /resources/:id/relationships/another-resource</tt>
@@ -232,18 +192,8 @@ module JSONAPI
       # * +source_record+ - The record whose relationship is modified
       # * +relationship_type+ - The relationship type
       def remove_to_one_relationship(source_record, relationship_type)
-        policy = ::Pundit.policy(user, source_record)
         relationship_method = "remove_#{relationship_type}?"
-        if policy.respond_to?(relationship_method)
-          unless policy.public_send(relationship_method)
-            raise ::Pundit::NotAuthorizedError,
-                  query: relationship_method,
-                  record: source_record,
-                  policy: policy
-          end
-        else
-          ::Pundit.authorize(user, source_record, 'update?')
-        end
+        authorize_relationship_operation(source_record, relationship_method)
       end
 
       # Any request including <tt>?include=other-resources</tt>
@@ -278,6 +228,22 @@ module JSONAPI
       # * +related_record+ - The associated record to return
       def include_has_one_resource(_source_record, related_record)
         ::Pundit.authorize(user, related_record, 'show?')
+      end
+
+      private
+
+      def authorize_relationship_operation(source_record, relationship_method, *args)
+        policy = ::Pundit.policy(user, source_record)
+        if policy.respond_to?(relationship_method)
+          unless policy.public_send(relationship_method, *args)
+            raise ::Pundit::NotAuthorizedError,
+                  query: relationship_method,
+                  record: source_record,
+                  policy: policy
+          end
+        else
+          ::Pundit.authorize(user, source_record, 'update?')
+        end
       end
     end
   end
--- a/lib/jsonapi/authorization/default_pundit_authorizer.rb
+++ b/lib/jsonapi/authorization/default_pundit_authorizer.rb
@@ -167,10 +167,16 @@ module JSONAPI
       # * +source_record+ - The record whose relationship is modified
       # * +new_related_records+ - The new records replacing the entire +has_many+
       #   association
-      #--
-      # TODO: Should probably take old records as well
-      def replace_to_many_relationship(_source_record, _new_related_records)
-        raise NotImplementedError
+      # * +relationship_type+ - The relationship type
+      def replace_to_many_relationship(source_record, new_related_records, relationship_type)
+        policy = ::Pundit.policy(user, source_record)
+        relationship_method = "replace_#{relationship_type}?"
+        allowed = if policy.respond_to?(relationship_method)
+                    policy.send(relationship_method, new_related_records)
+                  else
+                    policy.update?
+                  end
+        raise ::Pundit::NotAuthorizedError unless allowed
       end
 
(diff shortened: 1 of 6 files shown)


matthias-g:
Given the following scenario:

```ruby
class User < ActiveRecord::Base
  has_and_belongs_to_many :projects
end

class Project < ActiveRecord::Base
  has_and_belongs_to_many :users
end
```

and 

```ruby
class UserResource < BaseResource
  has_many :projects
end

class ProjectResource < BaseResource
  has_many :users
end
```

A user can be added to a project through
`POST /projects/1/relationships/users`
```
"data": [
  {
    "type": "users",
    "id": 1
  }
]
```

Now you can allow this action independently from allowing to update the project using `allow_relationship_users?`:
```ruby
class ProjectPolicy < ApplicationPolicy
  def allow_relationship_users? users
    true
  end

  # but
  def update?
    false
  end
end
```

valscion:
Yeah, that could work to allow for the relationship operations first to define more specific authorizations 🤔. That way we could first implement the relationship operations and only afterwards think about implementing the same thing in the generic update.

I think I was too bogged down in thinking how to allow `PATCH` calls directly to `/projects/1` URL and didn't think of how it would look like for the relationship operations specifically.

We'd need tests to avoid breaking this functionality in the future and to make sure it works properly. The current tests are a bit... unwieldy right now, sorry about that 😳. They are meant to be request-level tests testing the actual JSON API payloads and ensuring authorization is ran properly and that all cases are covered.

Is there anything I could help you with?

lib/jsonapi/authorization/default_pundit_authorizer.rb
```diff
@@ -146,8 +146,16 @@ def replace_to_one_relationship(_source_record, _old_related_record, _new_relate
...
-        raise NotImplementedError
+      # * +relationship_type+ - The relationship type
+      def create_to_many_relationship(source_record, new_related_records, relationship_type)
+        policy = Pundit.policy(user, source_record)
+        relationship_method = 'allow_relationship_' + relationship_type.to_s + '?'
+        if policy.respond_to?(relationship_method)
+          allowed = policy.send(relationship_method, new_related_records)
+        else
+          allowed = policy.update?
+        end
```
valscion: The build is currently failing because [Rubocop](https://github.com/bbatsov/rubocop) would like this piece of code to look something like this:

```rb
allowed =
  if policy.respond_to?(relationship_method)
    policy.send(relationship_method, new_related_records)
  else
    policy.update?
  end
```

matthias-g:
Thanks for the feedback! I've just updated the code according to your reviews and added some tests.

valscion:
Thanks! Do you want to keep on going with this approach for the other relationship operation types, too? I think we could take this as the first proper approach for handling the complex cases and then later look into what it would mean if we'd get it done for the common `PATCH /comments/1` case, too.

If you do want to keep on going for the other relationship types, too, I guess you could modify the pull request title to really say how amazing things are happening here :smile:

valscion:
That is, what do you want the scope of this pull request to be? Do you think this PR should be merged at some point, and if so, when do you feel like this PR is done? What things would you like to see implemented here?

matthias-g:
Yeah, I think the other relationship operation types should be implemented as well before the PR is mergeable. I'd say that the result of this PR should be that as a user of JA you can define policies for relationship operations on a per relationship level (instead of relying just on `update?`).

I'm still in the process of trying out though and I don't have a clear picture of how the result should look like. As a next step I've implemented the `remove_to_many_relationship` method.
I think this should use another policy method than `create_to_many_relationship` because a user might be allowed to create a relationship but not to delete it or the other way around. In the example above a user might not be allowed to add himself to a project but to delete himself from a project.
I've called the new method `remove_from_<relationship_name>?`. What do you think about renaming `allow_relationship_<relationship_name>?` to `add_to_<relationship_name>?`? To me that would seem more intuitive.

What I noticed and wanted to document here is that a user of JA needs to ensure that `allow_relationship_users?` and `allow_relationships_projects?` return the same results. This is not done using the current implementation and I don't see a good way to do this in JA but them returning different results could lead to unwanted behaviour (in case `POST /projects/1/relationships/users` is forbidden but `POST /users/1/relationships/projects` is allowed although they do the same).

valscion:
> Yeah, I think the other relationship operation types should be implemented as well before the PR is mergeable. I'd say that the result of this PR should be that as a user of JA you can define policies for relationship operations on a per relationship level (instead of relying just on update?).

:clap: Yes, great that you think so, too! I'll modify the PR title accordingly.

I think that we can leave out the entire `update?` method for now and figure that out later, if at all possible. We should be able to reuse the same policy methods for the added/removed associations in there later, too.

> I'm still in the process of trying out though and I don't have a clear picture of how the result should look like

Yeah I figured that only after testing out actually implementing these, would we discover things that cannot easily be seen upfront. The final solution can be something different than the original proposal idea, as long as the principles I outlined back in #30 stand:

**Prefer being overly restrictive rather than too permissive by accident.**

I really appreciate you communicating your ideas and I'm very open in rethinking a solution together 😌 



> As a next step I've implemented the remove_to_many_relationship method.
> I think this should use another policy method than create_to_many_relationship because a user might be allowed to create a relationship but not to delete it or the other way around.

:+1: Sounds reasonable, and this gem should allow for such a scenario to be handled, too.

> I've called the new method `remove_from_<relationship_name>?`

I like that name :+1:



> What do you think about renaming allow_relationship_<relationship_name>? to add_to_<relationship_name>?? To me that would seem more intuitive.

:+1: This makes sense, feel free to do so. That has a nice symmetry with the remove method.



> What I noticed and wanted to document here is that a user of JA needs to ensure that allow_relationship_users? and allow_relationships_projects? return the same results. This is not done using the current implementation and I don't see a good way to do this in JA but them returning different results could lead to unwanted behaviour (in case POST /projects/1/relationships/users is forbidden but POST /users/1/relationships/projects is allowed although they do the same).

Yes, this was in fact a big thing discussed over at #30 and was one of the reasons why the discussion got a bit heated.

There were talk about inspecting the relationship itself and then, with some logic, decide which of the two policies should be called. I liked the idea @aaronbhansen posted in https://github.com/venuu/jsonapi-authorization/issues/30#issuecomment-238598174, but I can't yet really form a proper opinion on the validity of that approach before I see some code implementing that feature.

Anyway, I'm a bit wary of going down that rabbit hole _yet_. Do you think it could be reasonable to solve that "both policies need to look the same" issue _after_ we've got the primitive API working in this PR? I fear that tackling that issue might open a can of worms where we must start to add swathes of tests to cover different types of relationships. And oh boy, there are many. It might be too soon to start figuring it out in this PR already.

I did hack around a bit to see if that would even remotely be possible, and I found out that yes, `jsonapi-resources` does have the information of relationships on the level that would make solving this double-the-logic-across-two-policies problem. I commented on it at https://github.com/venuu/jsonapi-authorization/issues/30#issuecomment-250911834:

> I haven't yet decided on which side of the association should handle the authorization. I'll have to restructure the spec dummy application a lot to include different kinds of relations so that I can be sure each case is tested properly 💦.
> 
> Based on my initial testings, it should be possible to get a nice approach for these authorization checks. Relationship objects coming from JR do know who owns the foreign key and allows jumping between the associated resources with relative ease.
> 
> This just might be quite big of an effort...

---

I'm really thankful for your efforts here! Would you like to get added as a contributor to this project once we get this out? ☺️

valscion:
Hiya @matthias-g! What's the status with this PR? Anything I can do to help you move forwards?

matthias-g:
Hey, thanks for asking. I'm sorry for being quiet the last weeks - I've just been busy with other work.
But I've already started this week working on the 'replace_to_many' and I think I'll be able to write the tests for that and commit them tomorrow. And to reply to your comment above. ;-)

valscion:
Yay, great to hear! If you need help bouncing around ideas or anything, we can book a https://room.co/ video chat (just open your browser and it works, no software install required) to talk things through if that would help not getting stuck with this :)

matthias-g:
> Anyway, I'm a bit wary of going down that rabbit hole yet. Do you think it could be reasonable to solve that "both policies need to look the same" issue after we've got the primitive API working in this PR?

Definitely. I'm not even sure if solving this on the side of JA is the best way. Maybe a design pattern for the policies could work as well. Like creating a concern per relationship that includes methods which are being called by the respective policy methods. But it's a good idea to discuss that later. :wink: 

> I'm really grateful for your efforts here! Would you like to get added as a contributor to this project once we get this out?

Thanks, that would be nice - once this has been merged.

---

For the recent commits I want to document two thoughts:
1. My first idea was to avoid creating a new policy method `replace_users?` by reusing `remove_from_users?` and `add_to_users?` instead (calling `remove_from_users?` for all old users and calling `add_to_users?` for the new users). But this doesn't work well enough as both methods are being called before any changes happen, meaning that `add_to_users?` is being called before the old users have been removed. This is a problem in the following scenario:
`add_to_users?` only returns true if the given users are not yet in the users relation. User A is in users and `replace_to_many` is called with new users including user A.
This would fail because user A can't be added, although the replace action is valid. That's why I used a new `replace_users?` method.

2. There was a TODO at `replace_to_many_relationship` in the authorizer: "Should probably take old records as well"
I've deleted this because the old records are available for the policy through source_record.relationship_name (like project.users). So it shouldn't be necessary to pass them to the policy.

valscion:
> But it's a good idea to discuss that later. 😉

:+1: Glad we're on the same page on this one 😅

> Thanks, that would be nice - once this has been merged.

We'll be glad to have you with us ☺️ 

I really like the way you went with the  `replace_users?` type of policy method. I also figured out the TODO comment being obsolete now as I read your commit and was happy to see you explain it more :smile:. It does feel a bit cleaner to be able to look at the old relationship models in the policy via normal means, just like with any other model. Love it :+1:.

Also, it makes sense to provide a separate `replace_X?` policy method. For example, it is usually desirable to prevent people from replacing all the comments on a `Post` and only allow adding to them. Having a separate policy method for the replace case seems like a good choice :+1:

matthias-g:
So, I've just pushed the implementations for the missing `replace_to_one` and `remove_to_one`.
To prevent inconsistencies I've changed the behavior slightly to calling `replace_X?` only in case an actual replacement happens. If you replace with `null` `remove_X?` is being called.

valscion:
:+1: Good call

valscion:
Huh, sorry, I hadn't figured out that this is already functionally complete! Sorry 😅 

Would you like to try out documenting this in the README somehow?

matthias-g:
> Huh, sorry, I hadn't figured out that this is already functionally complete! Sorry :sweat_smile:

No problem. I didn't stress that because I was still thinking about how the new policy methods could be useful for the generic update. But I didn't find a good way and I actually think that it is not that important as you can use the specific relationship operations if `update?` is too permissive.

> Would you like to try out documenting this in the README somehow?

Sure. I've just added a paragraph to the README. What do you think about it?

valscion:
This is looking super great, thank you! We'll look at this PR with @lime soon and reply back to you and maybe even merge this as it is 😊

valscion:
All in all this looks very good and we really appreciate your effort! We have some ideas on what we'd like to see here. I'll create a PR against your PR based on our review to get this shipped ☺️

valscion:
Here's some suggestions we came up with @lime as we reviewed this PR. What do you think?

- [ ] Add tests for every scenario of the relationship operations
  * Check out https://github.com/matthias-g/jsonapi-authorization/pull/1 — it points to this PR, so once merged this PR will update to contain its commits
- [ ] Add some helping comments about the fallbacking to `update?` in specs
  * Right now it isn't very obvious why these cases don't have the policy method defined
  * Maybe just a comment like this above each of the test cases could do?
    ```rb
    context 'where remove_<type>? not defined' do
      # CommentPolicy does not define #remove_article? method, so test using it
      subject(:method_call) do
        -> { authorizer.remove_to_one_relationship(source_record, :article) }
      end
    ```
- [ ] The `Pundit::NotAuthorizedError` is missing its positional arguments:
  1. `query` parameter, e.g. `add_to_comments?`
  2. `record`
  3. `policy`
  - These are used in the authorization error messages, so they should be set
- [ ] README.md could talk about many-to-many relationships instead of `M:N`
  * People are more familiar with the term many-to-many than `M:N`, I think
- [ ] The API of `DefaultPunditAuthorizer` methods should be changed to use keyword arguments
  * Currently the API change is backwards-incompatible in a way that could be very confusing,
    namely as `replace_to_one_relationship` has the same amount of parameters but are very
    different than what there was before
  * Keyword arguments are supported ever since Ruby 2.0 so we can use them

Merging this could be the first step to getting to 1.0 release. It might be a good idea to create a "Roadmap to 1.0" issue that documents what needs to be done until we're happy with a major version release. Before that, though, we could release a `1.0.0-alpha.1` version soon-ish.

gnfisher:
It looks like including relationships in the payload on a POST or PATCH to a resource still just defaults to checking permissions via `update?` on the associated record?

I have done a bit of work  on these based on the work @valscion did in his `proto-finer-grained relationships branch` that I think can be tweaked to use the same `add_to_*`, 'replace_*` and `remove_*` style policy actions. That would cover all scenarios, I think.

gnfisher:
I am also curious what is thought to be a better way of handling has_many related records: sending them as an array to the relationship method (like in this PR) or iterating through the array of records in the DefaultPunditAuthorizer class and passing one record at a time to the relationship method?

Basically, `add_to_users?(users)` or `add_to_users?(user)`?

matthias-g:
Thanks for your constructive feedback, @valscion, @lime!

> Add some helping comments about the fallbacking to update? in specs

Comments added. :heavy_check_mark:

> The Pundit::NotAuthorizedError is missing its positional arguments

Thanks, I had overlooked that. :heavy_check_mark:

> README.md could talk about many-to-many relationships instead of `M:N`

It does now. :heavy_check_mark:

matthias-g:
@gnfisher Thanks for reviewing and providing feedback as well!

> It looks like including relationships in the payload on a POST or PATCH to a resource still just defaults to checking permissions via update? on the associated record?

Yes, that is true. And I prefer an implementation that uses the relationship policy actions, too.
However I'd suggest to put this in a separate pull request, because this has potential for much discussion (as you can see in #30 :wink:). 

> The remove_to_one_relationship callback doesn't pass the record to the policy method. In my apps situation, where a Group has a Parent Group, it would be ideal to have access to the record being removed as well, so I can check the user has permission to modify that parent, too.

This is because the other relationship callbacks don't pass the current related records as well.
Isn't it possible to access the Parent Group through something like `record.parent` in your GroupPolicy?

> Basically, `add_to_users?(users)` or `add_to_users?(user)`?

That is a good question. And the current implementation is even inconsistent: It's `add_to_comments?(comments)` but `remove_from_comments?(comment)`.
Currently I don't see that one option is better. So I decided to copy the behavior of JR. There the inconsistency will be [fixed in version 0.9](https://github.com/cerebris/jsonapi-resources/commit/78e3560b84f3f99942f86b9a45e3c3c494295752).

gnfisher:
> This is because the other relationship callbacks don't pass the current related records as well.
Isn't it possible to access the Parent Group through something like record.parent in your GroupPolicy?

Of course. :man_facepalming:  Sorry!

> That is a good question. And the current implementation is even inconsistent: It's add_to_comments?(comments) but remove_from_comments?(comment).
Currently I don't see that one option is better. So I decided to copy the behavior of JR. There the inconsistency will be fixed in version 0.9.

Ok. yes, I spoke to Larry in the JR chat about some other topic and I do remember him saying that in the next release it will pass all records versuses going through the processor for every record in the array. Awesome!

valscion:
Looks great to me, thank you! I will follow-up later with more PRs and the 1.0 roadmap issue until releasing a new version of this gem.

I expect to have time for this next week. We are tracking progress on getting this done on our internal Kanban board at work so this shouldn't be forgotten :smile:

PR #47, opened 2017-02-20
Add tests for every scenario of the relationship operations

valscion:
We want to make sure the tests for the new cases match all the possible cases:
  * authorized with finer grained policy but NOT authorized with `update?`
  * NOT authorized with finer grained policy but authorized with `update?`
  * authorized with both finer grained policy and `update?`
  * :point_up: This also means that the current tests could be merged together

I needed to do some refactoring with the stubbing logic to make it work correctly for all these cases.

### DIFF
--- a/spec/jsonapi/authorization/default_pundit_authorizer_spec.rb
+++ b/spec/jsonapi/authorization/default_pundit_authorizer_spec.rb
@@ -302,7 +302,7 @@ RSpec.describe JSONAPI::Authorization::DefaultPunditAuthorizer do
       it { is_expected.not_to raise_error }
     end
 
-    context 'unauthorized for update? on record' do
+    context 'unauthorized for replace_author? and update? on record' do
       before do
         stub_policy_actions(source_record, replace_author?: false, update?: false)
       end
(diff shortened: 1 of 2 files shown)


valscion:
Review from @lime or @matthias-g would be appreciated ☺️

I added you as a collaborator now, @matthias-g :tada:. All pull requests will need a review before merging, so if you want to help you can review open pull requests, too, and give comments to help the PRs advance.

You'll notice that we now also have a few labels — these are similar to what we use internally at Venuu, so let me walk through them quickly:

## Type labels

These labels allow us to distinguish issues and PRs from each other. It's OK if a PR or an issue doesn't have any type label set.

Type: Discussion

* Applied to issues that are actually discussions

Type: Bug

* Applied to issues that are bugs (something is broken) and PRs that fix bugs.

Type: Critical bug

* Applied to issues and PRs that fix severe bugs.
* This label should not be added lightly — it is something like "THE SHIP IS ON FIRE" and should only be used if something very severe is wrong.

## Status labels

These are labels that would be set by the PR author, if it would be possible. They allow us to see when the PR is in need of a review or other help.

Help: Input appreciated
  
* Someone has trouble continuing with a PR and needs outside comments to continue.
* After the question is resolved and work can continue, this label should be removed.
* We try to answer cases with this label quickly

Status: Ready for review

* A pull request is ready for reviewing and waits for a collaborator to review it

Status: PR / Help wanted

* Applied to issues that would accept a PR closing the issue.

## Review labels

These labels allow us to quickly see what the review status of a PR is. These are set by the _reviewer_ to show the review status.

Review: Feedback given

* Pull request has got a review and needs some further work to be done before it can be merged. The PR has comments about the actions needed

Review: Ready to ship

* Pull request is accepted for merging

valscion:
Thanks for the review :)

PR #51, opened 2017-03-08
Handling relationship policy checking in replace_fields

gnfisher:
FYI: This breaks tests. I haven't had time to fix them, but wanted to just put this code out to start discussion in the mean time.

This will check permissions on new records being added. It uses the same method syntax as @matthias-g's #40 recently merged PR. i.e. `add_comment`, `remove_comment`, `add_to_users`, `remove_from_users`, etc. It also creates a new method syntax for polymorphic source records. 

For example, if comments are polymorphic, in the comment policy you could check if the user has permissions to comment on an article with `add_commentable_article?(article)`. This would cover payloads like:
```
# POST api/v1/comment
data: {
  type: 'comment,
  attributes: {
    ...
  },
  relationships: {
    commentable: {
      data: { type: 'articles', id: 1 }
      }
    }
...
}
```

Also, the `remove_to_one_relationship` method was not updated to match the rest of the endpoints in @matthias-g 's branch - so it seemed to me - and I updated it here.

Again, need to update the test suite to handle this way of doing things and do more tests around polymorphic scenarios, etc. But it's something to start more conversation on perhaps.

Also this is my first non trivial code contribution to an OSS project so please let me know how I can best help out with this! This is the code I was offering to share in issue #30.

### DIFF
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -210,8 +210,16 @@ module JSONAPI
           params[:resource_id],
           context: context
         )._model
-
+        
         relationship_type = params[:relationship_type].to_sym
+        relationship_resource = @resource_klass
+          ._relationship(relationship_type)
+          .resource_klass
+          .find_by_key(
+            params[:associated_key],
+            context: context
+          )
+        related_record = related_resource._model unless related_resource.nil?
 
         authorizer.remove_to_one_relationship(source_record, relationship_type)
       end
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -271,15 +271,16 @@ module JSONAPI
                 resource_class = @resource_klass.resource_for(assoc_value[:type].to_s)
                 resource_class.find_by_key(assoc_value[:id], context: context)._model
               when Array
-                resource_class = @resource_klass.resource_for(assoc_name.to_s)
+                resource_class = resource_class_for_relationship(assoc_name)
                 resource_class.find_by_keys(assoc_value, context: context).map(&:_model)
               else
                 resource_class = resource_class_for_relationship(assoc_name)
                 primary_key = resource_class._primary_key.to_sym
                 resource_class._model_class.find_by(primary_key => assoc_value)
               end
+
             {
-              relationship: @resource_klass._relationships[assoc_name],
+              relation_type: rel_type,
               relation_name: assoc_name,
               records: related_models
             }
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -269,6 +269,41 @@ module JSONAPI
         end
       end
 
+      def related_models_with_context
+        data = params[:data]
+        return { relationship: nil, relation_name: nil, records: nil } if data.nil?
+
+        [:to_one, :to_many].flat_map do |rel_type|
+          data[rel_type].flat_map do |assoc_name, assoc_value|
+            related_models =
+              case assoc_value
+              when Hash # polymorphic relationship
+                resource_class = @resource_klass.resource_for(assoc_value[:type].to_s)
+                resource_class.find_by_key(assoc_value[:id], context: context)._model
+              when Array
+                resource_class = @resource_klass.resource_for(assoc_name.to_s)
+                resource_class.find_by_keys(assoc_value, context: context).map(&:_model)
+              else
+                resource_class = resource_class_for_relationship(assoc_name)
+                primary_key = resource_class._primary_key.to_sym
+                resource_class._model_class.find_by(primary_key => assoc_value)
+              end
+            old_records = if params[:resource_id]
+                            @resource_klass.find_by_key(
+                              params[:resource_id],
+                              context: context
+                            )._model.send(assoc_name)
+                          end
+            {
+              relationship: @resource_klass._relationships[assoc_name],
+              relation_name: assoc_name,
+              records: related_models,
+              old_records: old_records
+            }
+          end
+        end
+      end
+
       def authorize_model_includes(source_record)
         if params[:include_directives]
           params[:include_directives].model_includes.each do |include_item|
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -103,18 +103,18 @@ module JSONAPI
       end
 
       def authorize_replace_fields
+        field_data = params[:data]
         source_record = @resource_klass.find_by_key(
           params[:resource_id],
           context: context
         )._model
-
-        authorizer.replace_fields(source_record, related_models)
+        authorizer.replace_fields(source_record, related_models_with_context)
       end
 
       def authorize_create_resource
-        source_class = @resource_klass._model_class
-
-        authorizer.create_resource(source_class, related_models)
+        field_data = params[:data]
+        source_class = resource_klass._model_class
+        authorizer.create_resource(source_class, related_models_with_context)
       end
 
       def authorize_remove_resource
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -216,12 +216,12 @@ module JSONAPI
           ._relationship(relationship_type)
           .resource_klass
           .find_by_key(
-            params[:associated_key],
+            params[:resource_id],
             context: context
           )
-        related_record = related_resource._model unless related_resource.nil?
+        related_record = relationship_resource._model unless relationship_resource.nil?
 
-        authorizer.remove_to_one_relationship(source_record, relationship_type)
+        authorizer.remove_to_one_relationship(source_record, related_record, relationship_type)
       end
 
       private
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -275,8 +275,7 @@ module JSONAPI
                 resource_class.find_by_keys(assoc_value, context: context).map(&:_model)
               else
                 resource_class = resource_class_for_relationship(assoc_name)
-                primary_key = resource_class._primary_key.to_sym
-                resource_class._model_class.find_by(primary_key => assoc_value)
+                resource_class.find_by_key(assoc_value[:id], context: context)._model
               end
 
             {
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -103,7 +103,6 @@ module JSONAPI
       end
 
       def authorize_replace_fields
-        field_data = params[:data]
         source_record = @resource_klass.find_by_key(
           params[:resource_id],
           context: context
@@ -112,7 +111,6 @@ module JSONAPI
       end
 
       def authorize_create_resource
-        field_data = params[:data]
         source_class = resource_klass._model_class
         authorizer.create_resource(source_class, related_models_with_context)
       end
@@ -283,7 +281,7 @@ module JSONAPI
             {
               relationship: @resource_klass._relationships[assoc_name],
               relation_name: assoc_name,
-              records: related_models,
+              records: related_models
             }
           end
         end
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -280,17 +280,10 @@ module JSONAPI
                 primary_key = resource_class._primary_key.to_sym
                 resource_class._model_class.find_by(primary_key => assoc_value)
               end
-            old_records = if params[:resource_id]
-                            @resource_klass.find_by_key(
-                              params[:resource_id],
-                              context: context
-                            )._model.send(assoc_name)
-                          end
             {
               relationship: @resource_klass._relationships[assoc_name],
               relation_name: assoc_name,
               records: related_models,
-              old_records: old_records
             }
           end
         end
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -210,18 +210,10 @@ module JSONAPI
           params[:resource_id],
           context: context
         )._model
-        
-        relationship_type = params[:relationship_type].to_sym
-        relationship_resource = @resource_klass
-          ._relationship(relationship_type)
-          .resource_klass
-          .find_by_key(
-            params[:resource_id],
-            context: context
-          )
-        related_record = relationship_resource._model unless relationship_resource.nil?
 
-        authorizer.remove_to_one_relationship(source_record, related_record, relationship_type)
+        relationship_type = params[:relationship_type].to_sym
+
+        authorizer.remove_to_one_relationship(source_record, relationship_type)
       end
 
       private
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -112,7 +112,7 @@ module JSONAPI
 
       def authorize_create_resource
         source_class = resource_klass._model_class
-        authorizer.create_resource(source_class, related_models_with_context)
+        authorizer.create_resource(source_class, related_models)
       end
 
       def authorize_remove_resource
--- a/lib/jsonapi/authorization/default_pundit_authorizer.rb
+++ b/lib/jsonapi/authorization/default_pundit_authorizer.rb
@@ -271,7 +271,7 @@ module JSONAPI
 
         case relationship
         when ->(rel) { rel.polymorphic }
-          polymorphic_type = data[:records].class.name.downcase
+          polymorphic_type = data[:records].class.name.underscore
           "#{prefix}_#{relationship.class_name.downcase}_#{polymorphic_type}?"
         when JSONAPI::Relationship::ToOne
           "#{prefix}_#{assoc_name}?"
--- a/lib/jsonapi/authorization/default_pundit_authorizer.rb
+++ b/lib/jsonapi/authorization/default_pundit_authorizer.rb
@@ -93,7 +93,25 @@ module JSONAPI
         related_records_with_context.each do |data|
           relationship  = data[:relationship]
           relation_name = data[:relation_name]
-          records   = data[:records]
+          records = data[:records]
+
+          # Authorize removing old records 
+          case relationship
+          when JSONAPI::Relationship::ToMany
+            old_records = source_record.send relation_name
+            authorize_relationship_operation(
+              source_record, 
+              relationship_method(data, prefix: 'remove'), 
+              old_records
+            )
+          else
+            authorize_relationship_operation(
+              source_record,
+              relationship_method(data, prefix: 'remove')
+            )
+          end
+
+          # Authorize adding new records
           authorize_relationship_operation(source_record, relationship_method(data), records)
         end
       end
@@ -246,18 +264,20 @@ module JSONAPI
         end
       end
 
-      def relationship_method(data)
+      def relationship_method(data, **options)
         relationship = data[:relationship]
         assoc_name   = data[:relation_name]
+        prefix       = options[:prefix] || 'add'
 
         case relationship
         when ->(relationship) { relationship.polymorphic }
           polymorphic_type = data[:records].class.name.downcase
-          "#add_#{relationship.class_name.downcase}_#{polymorphic_type}?"
+          "#{prefix}_#{relationship.class_name.downcase}_#{polymorphic_type}?"
         when JSONAPI::Relationship::ToOne
-          "add_#{assoc_name}?"
+          "#{prefix}_#{assoc_name}?"
         else
-          "add_to_#{assoc_name}?"
+          prefix_preposition = prefix == 'add' ? 'to' : 'from'
+          "#{prefix}_#{prefix_preposition}_#{assoc_name}?"
         end
       end
     end
--- a/lib/jsonapi/authorization/default_pundit_authorizer.rb
+++ b/lib/jsonapi/authorization/default_pundit_authorizer.rb
@@ -98,12 +98,15 @@ module JSONAPI
       # ==== Parameters
       #
       # * +source_class+ - The class of the record to be created
-      # * +related_records_with_context+ - A hash with the relationship,
-      # relation name, and records to be associated with the new record. This
-      # will contain the records specified in the "relationships" key in the request
-      def create_resource(source_class, related_records_with_context)
+      # * +related_records+ - An array of records to be associated to the new
+      #   record. This will contain the records specified in the
+      #   "relationships" key in the request
+      def create_resource(source_class, related_records)
         ::Pundit.authorize(user, source_class, 'create?')
-        authorize_related_records(source_class, related_records_with_context)
+
+        related_records.each do |record|
+          ::Pundit.authorize(user, record, 'update?')
+        end
       end
 
       # <tt>DELETE /resources/:id</tt>
--- a/lib/jsonapi/authorization/default_pundit_authorizer.rb
+++ b/lib/jsonapi/authorization/default_pundit_authorizer.rb
@@ -292,6 +292,23 @@ module JSONAPI
           ::Pundit.authorize(user, source_record, 'update?')
         end
       end
+
+      def relationship_method(data, **options)
+        relationship = data[:relationship]
+        assoc_name   = data[:relation_name]
+        prefix       = options[:prefix]
+
+        case relationship
+        when ->(relationship) { relationship.polymorphic }
+          polymorphic_type = data[:records].class.name.downcase
+          "#{prefix}_#{relationship.class_name.downcase}_#{polymorphic_type}?"
+        when ->(relationship) { relationship.class == JSONAPI::Relationship::ToOne }
+          "#{prefix}_#{assoc_name}?"
+        else
+          prefix_preposition = prefix == 'add' ? 'to' : 'from'
+          "#{prefix}_#{prefix_preposition}_#{assoc_name}?"
+        end
+      end
     end
   end
 end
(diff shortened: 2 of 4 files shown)


lib/jsonapi/authorization/default_pundit_authorizer.rb
```diff
@@ -191,9 +238,9 @@ def remove_to_many_relationship(source_record, related_record, relationship_type
       #
       # * +source_record+ - The record whose relationship is modified
       # * +relationship_type+ - The relationship type
-      def remove_to_one_relationship(source_record, relationship_type)
+      def remove_to_one_relationship(source_record, related_record, relationship_type)
         relationship_method = "remove_#{relationship_type}?"
-        authorize_relationship_operation(source_record, relationship_method)
+        authorize_relationship_operation(source_record, relationship_method, related_record)
```
valscion: Removing an association doesn't need the old records, as it will be available directly from the record inside the Pundit policies. Wasn't that what you discussed with @matthias-g in https://github.com/venuu/jsonapi-authorization/pull/40#issuecomment-280493499?

lib/jsonapi/authorization/default_pundit_authorizer.rb
```diff
@@ -104,14 +137,28 @@ def replace_fields(source_record, new_related_records)
...
+                    else
+                      policy.update?  
+                    end
+
+          unless allowed
+           raise ::Pundit::NotAuthorizedError,
+                   query: add_method_name,
+                   record: records,
+                   policy: policy
+          end
```
valscion: I think you could use the `authorize_relationship_operation` helper method a lot in here. It would hopefully cut down the duplication in here a lot :)

lib/jsonapi/authorization/default_pundit_authorizer.rb
```diff
@@ -245,6 +292,23 @@ def authorize_relationship_operation(source_record, relationship_method, *args)
...
+      def relationship_method(data, **options)
+        relationship = data[:relationship]
+        assoc_name   = data[:relation_name]
+        prefix       = options[:prefix]
+
+        case relationship
+        when ->(relationship) { relationship.polymorphic }
+          polymorphic_type = data[:records].class.name.downcase
+          "#{prefix}_#{relationship.class_name.downcase}_#{polymorphic_type}?"
+        when ->(relationship) { relationship.class == JSONAPI::Relationship::ToOne }
```
valscion: I think these case-statements could be written like so?

```rb
case relationship
when relationship.polymorphic
  # ...
when JSONAPI::Relationship::ToOne
  # ...
else
  # ...
end
```

I haven't seen this sort of lambda usage before myself 😅

valscion:
Is it so that you are now using the class name of the actual record to figure out the policy method name? If I understood correctly, in @matthias-g's PR we went with using the relation name from the resource class itself as-is instead of figuring out the class name behind the association.

I think if we'd do the same in here, we could make the code much more simpler and wouldn't have to worry about polymorphic associations as much?

valscion:
Thanks for the PR! This feature is what is missing indeed and we'd like to get it to the 1.0 release 😊 .

Let's figure out how to test this with the current test suite once we agree on a direction where this is going :)

gnfisher:
Just to be clear, we are not falling back to `update?` any longer?

lib/jsonapi/authorization/default_pundit_authorizer.rb
gnfisher: Ok, that makes sense. But in that case there is an inconsistency with `to_many` records, wouldn't `remove_to_many_relationship` have the same rules? Right now it takes an argument of related records.

lib/jsonapi/authorization/default_pundit_authorizer.rb
gnfisher: In order to call a method on the case object (like `.polymorphic` and `.class`) I have to do this via a lambda. `when relationship.polymorphic` and `relationship.class == JSONAPI::Relationship::HasOne` always come back false even when they should be true. Perhaps I am missing something here? Please let me know. It only works properly for me when using lambdas.

valscion: Yes, it seems to be so that you have to use lambda to check for the `relationshop.polymorphic` case. Ruby doesn't seize to surprise me :smile:

For the class case, though, you shouldn't need it to be a lambda.

screen shot 2017-03-13 at 10 00 41

valscion:
I think we will need to call the `remove_from_#{relationship}?` and `remove_#{relationship}?` methods for each record that there used to be when replacing fields.

@matthias-g, what do you think about the case when we're PATCHing away an entire relationship?

Here's the setup:

```rb
Article.new(id: 1, comments: [comment_with_id_1, comment_with_id_2]).save
```

And we'd send this JSON API request:

`PATCH /articles/1`

```
"data": {
  "type": "articles",
  "id": "1",
  "relationships": {
    "comments": {
      "data": [
        { "type": "comments", "id": "3" }
      ]
    }
  }
}
```

What authorization methods should such an operation use? If we didn't have any other methods besides the `add` and `remove` ones, I think it would have to be like this:

1. `ArticlePolicy#remove_from_comments(comment_with_id_1)`
2. `ArticlePolicy#remove_from_comments(comment_with_id_2)`
3. `ArticlePolicy#add_to_comments(comment_with_id_3)`

Would that make sense?

If we think this is the solution we should use, we'll need to have specs checking that these are the calls that are being made and authorized against.

valscion:
The build is currently failing because of code style issues:

```
lib/jsonapi/authorization/authorizing_processor.rb:106:9: W: Useless assignment to variable - field_data.
        field_data = params[:data]
        ^^^^^^^^^^
lib/jsonapi/authorization/authorizing_processor.rb:115:9: W: Useless assignment to variable - field_data.
        field_data = params[:data]
        ^^^^^^^^^^
lib/jsonapi/authorization/authorizing_processor.rb:286:38: C: Avoid comma after the last item of a hash.
              records: related_models,
                                     ^
lib/jsonapi/authorization/default_pundit_authorizer.rb:98:43: C: Trailing whitespace detected.
          # Authorize removing old records 
                                          ^
lib/jsonapi/authorization/default_pundit_authorizer.rb:103:29: C: Trailing whitespace detected.
              source_record, 
                            ^
lib/jsonapi/authorization/default_pundit_authorizer.rb:104:59: C: Trailing whitespace detected.
              relationship_method(data, prefix: 'remove'), 
                                                          ^
lib/jsonapi/authorization/default_pundit_authorizer.rb:124:73: C: Trailing whitespace detected.
      # * +related_records_with_context+ - A hash with the relationship, 
                                                                        ^
lib/jsonapi/authorization/default_pundit_authorizer.rb:273:17: W: Shadowing outer local variable - relationship.
        when ->(relationship) { relationship.polymorphic }
                ^^^^^^^^^^^^
```

I'm sorry that we don't have a nicer way of running the lint with CI yet 😞 . See #34 for some history regarding these lint failures.

gnfisher:
> Is it so that you are now using the class name of the actual record to figure out the policy method name? If I understood correctly, in @matthias-g's PR we went with using the relation name from the resource class itself as-is instead of figuring out the class name behind the association.
> 
> I think if we'd do the same in here, we could make the code much more simpler and wouldn't have to worry about polymorphic associations as much?

Sorry but I don't quite follow on this. What situation(s) are you talking about specifically?

gnfisher:
> What authorization methods should such an operation use? If we didn't have any other methods besides the add and remove ones, I think it would have to be like this:
> 
>     ArticlePolicy#remove_from_comments(comment_with_id_1)
>     ArticlePolicy#remove_from_comments(comment_with_id_2)
>     ArticlePolicy#add_to_comments(comment_with_id_3)
> 
> Would that make sense?

This is how it is handled in this branch.

valscion:
> Sorry but I don't quite follow on this. What situation(s) are you talking about specifically?

Hmm I guess I'm just a bit confused yet as to what Pundit policy method will be called when the resource specifies a custom `:relation_name` with the relationship http://jsonapi-resources.com/v0.8/guide/resources.html#Options

Here's the setup:

```rb
Comment.new(id: 1, reviewing_user: user_with_id_2).save
```

```rb
class CommentResource < JSONAPI::Resource
  include JSONAPI::Authorization::PunditScopedResource

  has_many :tags
  has_one :article
  has_one :reviewer, relation_name: "reviewing_user", class_name: "User"
end
```

And we'd send this JSON API request:

`PATCH /comments/1`

```json
"data": {
  "type": "articles",
  "id": "1",
  "relationships": {
    "reviewer": {
      "data": {
        "type": "user",
        "id": "3"
      }
    }
  }
}
```

Will we call `CommentPolicy#remove_reviewer?` or `CommentPolicy#remove_reviewing_user?` method? And the same for `#add_reviewer?` vs. `#add_reviewing_user?`.

I think we might be better off if we'd just pick the relation name from the relationship definition, not the `:relation_name` that was given. I.e. call `CommentPolicy#remove_reviewer?` and forget about the custom `:relation_name`.

What do you think?

gnfisher:
> Will we call CommentPolicy#remove_reviewer? or CommentPolicy#remove_reviewing_user? method? And the same for #add_reviewer? vs. #add_reviewing_user?.
> 
> I think we might be better off if we'd just pick the relation name from the relationship definition, not the :relation_name that was given. I.e. call CommentPolicy#remove_reviewer? and forget about the custom :relation_name.
> 
> What do you think?

Ok, I will investigate and report back.

matthias-g:
> What authorization methods should such an operation use? If we didn't have any other methods besides the add and remove ones, I think it would have to be like this:
> 
>     ArticlePolicy#remove_from_comments(comment_with_id_1)
>     ArticlePolicy#remove_from_comments(comment_with_id_2)
>     ArticlePolicy#add_to_comments(comment_with_id_3)
> 
> Would that make sense?

I agree.

But I see two things to consider:
  1. We should make sure that in case `data` contains also `{ "type": "comments", "id": "2"}` only `ArticlePolicy#remove_from_comments(comment_with_id_1)` and `ArticlePolicy#add_to_comments(comment_with_id_3)` are being called (instead of checking if `comment_with_id_2` can be removed and added immediately afterwards).
  2. The article record that is used when calling `add_to_comments` should already have comments 1 and 2 removed. In case the policy checks that the number of comments does not exceed 2 the request should be allowed.

valscion:
> The article record that is used when calling add_to_comments should already have comments 1 and 2 removed

I don't think we will ever be at that stage, considering the authorization happens before any changes to models are made.

I'm not quite sure what you're saying, so maybe you could elaborate this second part?

valscion:
Yeah, I think you're right, Matthias.

So I guess we should aim for a `replace_#{relationship}?` method, then? That seems to be the only method that would cover even the use case you described.

valscion:
By the way, at this point I want to thank you @gnfisher for starting work on this PR! It is super nice to be able to discuss these different implementations when we've got some working code under our belts :wink:

Also thanks to @matthias-g for joining the conversation. It's great to have you both here figuring out a good approach to this tricky problem.

I think we're heading to a good direction with this PR ☺️. Once we know what end result we'll want to have, I can help you figure out how to add some tests to cover this feature properly 😊

gnfisher:
Thank both you of you - JA has been a huge help at work and a fun learning opportunity for me. 

As for `replace_(relationship)?` I guess I just don't see the utility still? In @matthias-g 's scenario, it seems to me like the developer would just have to keep in mind when writing policies that authorization is done before records are modified. Really, that sort of logic (limiting number of comments) seems like more of a model validation than a authorization issue? What do you guys think?

valscion:
Consider which API allows you to do it, and which API doesn't allow you to build one.

We might have a use case for this in the future at Venuu. Only admins could be allowed to add more than 5 suitable `EventType` associations for a given `Venue`, to avoid people abusing our search engine.

It would be best if this could be handled at the authorization layer.

valscion:
And also, it would be fairly easy to just create a `replace_#{relationship}?` policy method that directly delegates to the `add_to_#{relationship}?` and `remove_from_#{relationship}?` methods. It is easily solved in the application level

gnfisher:
> Consider which API allows you to do it, and which API doesn't allow you to build one.
> 
> We might have a use case for this in the future at Venuu. Only admins could be allowed to add more than 5 suitable EventType associations for a given Venue, to avoid people abusing our search engine.
> 
> It would be best if this could be handled at the authorization layer.

Hm, okay. This is a good example. Thanks.

> And also, it would be fairly easy to just create a replace_#{relationship}? policy method that directly delegates to the add_to_#{relationship}? and remove_from_#{relationship}? methods. It is easily solved in the application level

So the idea then is that in the `replace_#{relationship}?` policy method the user would be able to add in a little more of their own logic and call the remove/add methods themselves?

If we handle call add/remove methods from within our `replace_*` method, it seems no different than what we do now?

Or would the idea be, call the `replace_#{relationship` authorization, then after returning true we would then call the `add_#{relationship}?` and `remove_#{relationship}` within out `replace_fields` action?

Would `replace_#{relationship}` only be called on `PATCH replace_fields` then?

valscion:
> So the idea then is that in the `replace_#{relationship}?` policy method the user would be able to add in a little more of their own logic and call the remove/add methods themselves?

Yup, that would be it.



> Or would the idea be, call the replace_#{relationship authorization, then after returning true we would then call the add_#{relationship}? and remove_#{relationship} within out replace_fields action?

Nono, we would call _only_ the `replace_#{relationship}?` method if it is defined, and then fall back to the current approach of using `#update?` on every record that would change, i.e. check both the new and old associated records.

> Would replace_#{relationship} only be called on PATCH replace_fields then?

Hmm, yes, as otherwise we can do with just using `add_to_#{relationship}?` and `add_#{relationship}?` methods when we get a POST request.

`replace_#{relationship}?` would be called also when we're doing a PATCH for has-many relationships operation, i.e. if we do a query for `PATCH /articles/1/relationships/comments`, we will call `ArticlePolicy#replace_comments?` method. This was how it was done in #40, [and here's a spec scenario testing it](https://github.com/venuu/jsonapi-authorization/blob/cb1d2434346f6e52fd149e89aa89e0d20c607db3/spec/requests/relationship_operations_spec.rb#L121-L181).

As you can see, when we use the `replace_#{relationship}?` method, we need to only pass the new records in to the policy method, as the old records can be queried from the model itself should it be necessary for the application authorization.

We also use the `replace_#{relationship}?` method for has-one relationship cases. Here's the [request spec that tests we're calling for `replace_to_one_relationship` authorization method](https://github.com/venuu/jsonapi-authorization/blob/cb1d2434346f6e52fd149e89aa89e0d20c607db3/spec/requests/relationship_operations_spec.rb#L183-L260) and here's the [authorization spec that tests we're using the `replace_#{relationship}?` method to test for authorization](https://github.com/venuu/jsonapi-authorization/blob/cb1d2434346f6e52fd149e89aa89e0d20c607db3/spec/jsonapi/authorization/default_pundit_authorizer_spec.rb#L289-L333)

I think we should try to mimic how relationship operations are authorized for and use the same checks for resource-level requests, too. So, these should be identical:

## `has-one` relationship

`PATCH /articles/1/relationships/author`

```json
{
  "data": {
    "type": "users",
    "id": "5"
  }
}
```

`PATCH /articles/1`

```json
{
  "type": "articles",
  "id": "1",
  "relationships": {
    "author": {
      "data": {
        "type": "users",
        "id": "5"
      }
    }
  }
}
```

## `has-many` relationship

`PATCH /articles/1/relationships/comments`

```json
{
  "data": [
    { "type": "comments", "id": "2" },
    { "type": "comments", "id": "3" }
  ]
}
```

`PATCH /articles/1`

```json
{
  "type": "articles",
  "id": "1",
  "relationships": {
    "comments": {
      "data": [
        { "type": "comments", "id": "2" },
        { "type": "comments", "id": "3" }
      ]
    }
  }
}
```

gnfisher:
I do remember the `replace_#{relationships}?` method in the original #40 from @matthias-g. Okay, this makes a lot of sense to me now. I like it. I guess in the README we'll just be explicit that in `replace_*` policy actions the user is responsible for add/remove checks etc? I believe @matthias-g original PR also had a note about just that.

Should I move forward with this refactor then?

valscion:
> Should I move forward with this refactor then?

Yes, please do :relaxed: :tada:. When in doubt on what to check for, the [specs for relationships operations](https://github.com/venuu/jsonapi-authorization/blob/cb1d2434346f6e52fd149e89aa89e0d20c607db3/spec/requests/relationship_operations_spec.rb#L78-L260) hopefully can provide you guidance on what we'd expect to test in each scenario 😌 

I think we can get a 1-to-1 similarity with the relationship requests authorizations that was done in #40. The only difference in this PR is that there can be multiple relationships that change in the same request, as relationship operation requests could always only do changes to relationship at a time.

valscion:
Hmm, I think you might be able to do all of this authorization by just calling the `replace_to_many_relationship` & friends methods from the `replace_fields` method! So you shouldn't need to figure out yourself what is the correct policy method to call, as the relationship operations already know how to handle it! :sparkles:

Here's an example scenario and how I image it would play out:

```rb
user_1 = User.new(id: 'user-1').save!
comment_1 = Comment.new(id: 'c-1', content: 'Is this the real life?').save!
comment_2 = Comment.new(id: 'c-2', content: 'Is this just fantasy?').save!
comment_3 = Comment.new(id: 'c-3', content: '(you know the rest)').save!
article_1 = Article.new(
              id: 'a-1',
              author: user_1,
              comments: [comment_1]
            ).save!
user_2 = User.new(id: 'user-2').save!
```

`PATCH /articles/a-1`

```json
{
  "type": "articles",
  "id": "a-1",
  "relationships": {
    "author": {
      "data": {
        "type": "users",
        "id": "user-2"
      }
    },
    "comments": {
      "data": [
        { "type": "comments", "id": "c-2" },
        { "type": "comments", "id": "c-3" }
      ]
    }
  }
}
```

1. Authorize `ArticlePolicy#update?`
2. Call `#replace_to_one_relationship(article_1, user_2, :author)`
3. Call `#replace_to_many_relationship(article_1, [comment_2, comment_3], :comments)`
4. That's it! There's no step 4!

Is there something I didn't catch with this?

valscion:
Hiya @gnfisher, anything I could do to help you move forwards with this?

gnfisher:
Sorry, I've had a difficult month time-wise! This weekend or early next week I hope to have some new code up here and push this towards completion.

valscion:
Great to hear that! I was just checking, thanks for the time you have put into this ☺️

gnfisher:
Using the existing helper methods (`replace_to_many_relationship` etc) works great and allows simplification.

The only issue is the following with polymorphic resources. 

For `Tag`/`Taggable` resources it looks for a `replace_taggable?(taggable)` method for authorizing in the `Tag` policy.

The only thing is, if a polymorphic resource is attached to many models, that `replace_taggable?` action could be pretty huge to handle all scenarios for all resources - imagine images, which could be attached to a dozen or more resources.

My old code expected `replace_taggable_<resource>?` type methods, which forces a bit cleaner organization of code.

Thoughts? 

I'm leaning towards the one-size-fits-all `replace_taggable?` way as it just keeps the JA code a bit leaner, and seems a little bit more straightforward for the end user. Right?

valscion:
> Using the existing helper methods (`replace_to_many_relationship` etc) works great and allows simplification.

Yess, great to hear that! :tada:

> The only issue is the following with polymorphic resources.

I agree with the one-size-fits-all approach you're leaning towards, too.

I think JA should at least first only lean on the simpler approach, as abstractions on top of `replace_taggable?(taggable)` method can be built in user land first. E.g. nothing is stopping users of JA to write their policies like this:

```rb
class TagPolicy
  def replace_taggable?(taggable)
    send("replace_taggable_#{taggable.class.to_s.underscore}?", taggable)
  end

  private

  def replace_taggable_comment?(comment)
    false
  end
end
```

If we later find out a particularly nice approach to this, we might be able to do it even outside of this gem and provide the necessary hooks for helping with writing these policies.

gnfisher:
I haven't gotten to working on the tests yet. Many fail, as the `create_resource` and `replace_fields` methods in the processor now take a has rather than an array of records, and the spec doubles don't like it very much :wink:.

I tested this on a demo app, however, and it handles everything well. If it looks good I can go about updating the tests and writing some new ones, but may need a little guidance. And also updating the readme with instructions regarding polymorphic relationships.

I will try to keep on this all week and get it done! Thanks for all your patience and assistance everyone.

valscion:
Hmm I think you might need to also call `create_to_many_relationship?` and `create_to_one_relationship?` authorizer methods when you're POSTing a new resource? Or am I mistaken?

gnfisher:
My understanding is that all of that logic goes inside of the `replace_to_one_relationship` action. From the docs:

```
 def replace_comments?(new_comments)
    allowed = record.comments.all? { |comment| new_comments.include?(comment) || add_to_comments?([comment])}
    allowed && new_comments.all? { |comment| record.comments.include?(comment) || remove_from_comments?(comment) }
  end
```

valscion:
Mm but in the new resource case, we don't yet have a `record` available in the Pundit policies as all we've got is the class.

```rb
user_1 = User.new(id: 'user-1').save!
comment_1 = Comment.new(id: 'c-1', content: 'Is this the real life?').save!
comment_2 = Comment.new(id: 'c-2', content: 'Is this just fantasy?').save!
```

`POST /articles`

```json
{
  "type": "articles",
  "relationships": {
    "author": {
      "data": {
        "type": "users",
        "id": "user-1"
      }
    },
    "comments": {
      "data": [
        { "type": "comments", "id": "c-1" },
        { "type": "comments", "id": "c-2" }
      ]
    }
  }
}
```

This will need to be addressed somehow... I'm afraid that we don't actually have any methods we could use for this case, though, as all of the relationship operation methods expect that we already have a record instance! Damn.

Should we call these methods in this case?

1. `ArticlePolicy#create?`
2. `ArticlePolicy#create_with_user?(user_1)`
3. `ArticlePolicy#create_with_comments?([comment_1, comment_2])`

I'm afraid that we haven't actually discussed this case at all. 😕 

What should we authorize in this case? 😧

valscion:
The case in the above comment might need new methods to handle the creation of resources because this is basically how the `ArticlePolicy` will be instantiated when we're creating new resources instead of modifying existing ones:

```rb
policy = ArticlePolicy.new(current_user, Article)
```

Note that we don't yet have an instance of `Article` model yet at this point, and we cannot have it.

gnfisher:
I don't think a new action is necessarily required but it is an option.

```ruby
# ArticlePolicy.rb

def initialize(user, record)
  @user = user
  @record = record.is_a? Class ? record.new() || record
end

def replace_comments?(new_comments)
  remove_from_comments?(record.comments) unless record.is_a? Class
  add_to_comments?(new_comments)
end
```
Could we just do this?

valscion:
> Could we just do this?

I'm afraid not 😞

We could however modify the scope of this PR to just handle the `PATCH` operation, though, and figure out what to do with creating resources and authorizing their relationships separately, in order to get this PR merged and to give room to figure this out on its own.

gnfisher:
Ah, I suppose an new instance would respond with an empty array and the `remove_from_comments?` action might return false? But couldn't you just `return true if new_comments.empty?` from those type of methods? I am not sure why setting record to a new instance is not a viable solution. What am I missing? :)

Ok, I have no problem with reducing the scope to just `replace_fields`.

valscion:
> But couldn't you just return true if new_comments.empty? from those type of methods?

Yes, but we don't want to put that burden on our users. Nobody likes the `NoMethodError: undefined method X for nil:NilClass`, and it would be surprising to get one in this case. That most likely would be because of a badly designed API.

gnfisher:
Ok I figured that was the reasoning. I will clean out `create_resources` and try to take care of tests on the `replace_fields` then. Meanwhile, we can open a new issue to specifically chew on the `create_resource` options?

valscion:
> Meanwhile, we can open a new issue to specifically chew on the create_resource options?

Done, see #56 :)

gnfisher:
- Returned create_resource to its previous state.
- Add spec coverage around `#replace_fields`

One failing  spec seems to me to be an issue with the instance double... I'm not quite sure how to resolve it. It's this spec, which is pending anyway, but still.

```ruby
# ./spec/requests/tricky_operations_spec.rb:132

      context 'limited by Comments policy scope' do
        let(:comments_policy_scope) { Comment.where("id NOT IN (?)", new_comments.map(&:id)) }
        before { allow_operation('replace_fields', article, related_records_with_context) }

        it do
          pending 'DISCUSS: Should this error out somehow?'
          is_expected.to be_not_found
        end
      end
```

lib/jsonapi/authorization/authorizing_processor.rb
```diff
@@ -261,6 +259,35 @@ def related_models
...
+              when Hash # polymorphic relationship
+                resource_class = @resource_klass.resource_for(assoc_value[:type].to_s)
+                resource_class.find_by_key(assoc_value[:id], context: context)._model
+              when Array
+                resource_class = resource_class_for_relationship(assoc_name)
+                resource_class.find_by_keys(assoc_value, context: context).map(&:_model)
+              else
+                resource_class = resource_class_for_relationship(assoc_name)
+                primary_key = resource_class._primary_key.to_sym
+                resource_class._model_class.find_by(primary_key => assoc_value)
```
gnfisher: No, that was existing code from #related_models I believe that i co-opted.

All changes made.

valscion:
Looking great! Do you think this would benefit from having the same specs for the case when the policy classes don't specify the `replace_<type>?` methods? That is, cases similar to these:

https://github.com/venuu/jsonapi-authorization/blob/master/spec/jsonapi/authorization/default_pundit_authorizer_spec.rb#L315-L332

```rb
    context 'where replace_<type>? not defined' do
      # CommentPolicy does not define #replace_article?, so #update? should determine authorization
      let(:related_records) { Article.new }
      subject(:method_call) do
        -> { authorizer.replace_to_one_relationship(source_record, related_record, :article) }
      end

      context 'authorized for update? on record' do
        before { allow_action(source_record, 'update?') }
        it { is_expected.not_to raise_error }
      end

      context 'unauthorized for update? on record' do
        before { disallow_action(source_record, 'update?') }
        it { is_expected.to raise_error(::Pundit::NotAuthorizedError) }
      end
    end
```

We can do that in another PR, though, as this is getting quite long. Thanks for your help!

gnfisher:
Absolutely, I will file a new PR to add them. And then later this week leave my thoughts on `#create_resource` discussion.

Thank you @valscion for your incredible patience and helping me to get something contributable finished. I learned a lot!

valscion:
Thank you for enduring all the way through this long PR. It's been a pleasure seeing you learn new things and being able to contribute to this project ☺️

Issue #36, opened 2016-10-01
Not compatible with v0.9.0 of jsonapi-resources


valscion:
I believe that changes done in https://github.com/cerebris/jsonapi-resources/pull/845 will be incompatible with what we're running with in here.

Adding an issue so it won't come as a surprise to us.

valscion:
> Any plan to support the new version of jsonapi-resources, are there some issues preventing this gem to be compatible yet?

Changes in https://github.com/cerebris/jsonapi-resources/pull/845 changed the relationship operation names, and thus breaks our callbacks, too.

> We would like to enable the new caching feature of 0.9 but we cannot upgrade because of the dependency lock on 0.8.
> 
> How could we help? Thanks!

If you could look into what kind of a version check we'd need to implement to support 0.8 and 0.9 versions of JR and submit a PR, that would be great ☺️. We'd need to make sure that Travis runs successfully on both 0.8 version and v0.9.0.beta3 version of `jsonapi-resources` for the PR to be acceptable.

We're not yet testing out 0.9 version of JR as it's beta, but a PR to support both 0.8 and 0.9 would be appreciated :+1:

valscion:
We'd like to have help tackling this issue! If you want to start tackling this, please leave a comment so others won't waste effort on the same task ☺️

We'd want to get this compatibility fixed for v1.0.0 release, if at all possible.

d4rky-pl:
If #52 was merged, does that mean this issue is already solved? :)

valscion:
Oh yeah, good point. I thought that I had closed this issue already 😅 .

PR #52, opened 2017-03-17
Make compatible with jsonapi-resources 0.9

handlers:
Hi! In an effort to make jsonapi-authorization's test suite pass against jsonapi-resources 0.9, this PR does a few things:

1. Conditionally sets callbacks for `:create_to_many_relationship` `:replace_to_many_relationship` `:remove_to_many_relationship` based on whether or not `JSONAPI::Processor` responds to the singular or plural version of the method (See https://github.com/venuu/jsonapi-authorization/issues/36).
2. Adds a callback and stub authorization method for `replace_polymorphic_to_one_relationship`, which is [present in jsonapi-resources 0.9](https://github.com/cerebris/jsonapi-resources/blob/v0.9.0/lib/jsonapi/processor.rb#L13) and [0.8.0.beta1](https://github.com/cerebris/jsonapi-resources/blob/v0.8.0.beta1/lib/jsonapi/processor.rb#L13). I don't think this is strictly necessary for 0.9 compatibility but is good to have in my opinion.
3. Tries to fix up `authorize_remove_to_many_relationship` by updating the logic to find all related records and call `authorizer.remove_to_many_relationship` on each of them.

When this branch is run against the suite with jsonapi-resources at 0.8.0.beta1, there are a [couple of failures](https://gist.github.com/handlers/3b0672cfe7daebb0bff465d12fb80e87). I am not sure what the backwards compatibility requirements are between point releases of the gem, but I definitely tried to stay somewhat backwards compatible in the conditional callback setting strategy.

I am super new to this project and definitely want to get some experienced 👀 on this! Happy to chop it up into separate PRs and fix stuff up, as this one has a few distinct threads to it.


valscion:
Hi, and welcome! Thank you for starting a PR for this. I'm very excited to soon jump to review it!

Hopefully it's ok that this is going to take me a few days at least until I can look into this more. I've got a tight schedule at work for a while, and as my wrist is aching again I won't be able to look at this on my freetime anytime soon.

I won't forget about this PR, though :). I'll check this out when I've got a good timeslot for this 💞

valscion:
Closed and reopened to try to trigger a CI build... wonder what's going on 😕

valscion:
This seems to be going to a good direction, thank you!

I'm not sure whether we've got a good answer to the backwards compatibility here. The relationship processor hooks have changed a bit, and when we started to authorize for relationship actions we hit these changes, too: https://github.com/venuu/jsonapi-authorization/pull/40#issuecomment-280493499

I'd be OK with going by just supporting 0.9 in a future release, as it would make the relationship authorizations cleaner as well. 0.9.0 is out of beta so it's fine if we only target that version, too.

Would you be OK with setting this PR up to only support 0.9 and removing the then-unnecessary checks from your changes?

handlers:
Yes, totally ok with the route of 0.9-only compatibility! Should make the
code a lot cleaner, too! 🙂

handlers:
OK, I have updated the PR to require jsonapi-resources 0.9 and updated the code accordingly. I also had to update the `remove_to_many_relationship](https://github.com/venuu/jsonapi-authorization/pull/52/files#diff-692452b50fa5d62bfcae1b833e3e3fb4R208) calls to match the [new method signature](https://github.com/venuu/jsonapi-authorization/blob/master/lib/jsonapi/authorization/default_pundit_authorizer.rb#L181).

spec/requests/relationship_operations_spec.rb
```diff
@@ -282,7 +282,7 @@
 
     context 'unauthorized for remove_to_many_relationship' do
       before do
-        disallow_operation('remove_to_many_relationship', article, kind_of(Comment), :comments)
+        disallow_operation('remove_to_many_relationship', article, kind_of(Comment), "comments")
```
valscion: Hmm I wonder whether this change was necessary?

lib/jsonapi/authorization/authorizing_processor.rb
```diff
@@ -216,6 +224,10 @@ def authorize_remove_to_one_relationship
         authorizer.remove_to_one_relationship(source_record, relationship_type)
       end
 
+      def authorize_replace_polymorphic_to_one_relationship
+        raise NotImplementedError
```
valscion: Good call with just raising `NotImplementedError` here.

We'll need to figure out how to authorize for this polymorphic case before shipping v1.0.0 but that might be better to discuss outside of this PR.

spec/dummy/app/resources/article_resource.rb
```diff
@@ -24,4 +24,8 @@ def id=(external_id)
   # Setting this attribute is an easy way to test that authorizations work even
   # when the model has validation errors
   attributes :blank_value
+
+  def self.creatable_fields(context)
+    super + [:id]
+  end
```
valscion: Hmm what was the reason for this change? Is there a reason why it is needed?

lib/jsonapi/authorization/authorizing_processor.rb
```diff
@@ -181,28 +182,35 @@ def authorize_replace_to_many_relationship
...
+        else
+          related_records = related_resources.map { |resource| resource._model }
+        end
+
+        related_records.each do |related_record|
+          authorizer.remove_to_many_relationship(
+            source_record,
+            related_record,
+            params[:relationship_type]
+          )
```
valscion: Instead of looping through all the `related_records`, we might want to instead change the API of `remove_to_many_relationship` to take an array of `related_records` (instead of one at a time).

That way we wouldn't even have to do the `if related_resource.empty?` check you're doing before calling the authorizer right now, too.

There was a comment https://github.com/venuu/jsonapi-authorization/pull/40#issuecomment-280493499 that maybe related to this not being possible before 0.9 version of JR.

valscion:
You'll want to change the `.travis.yml` to configure Travis to use `0.9.0` version of JR when running tests

lib/jsonapi/authorization/authorizing_processor.rb
```diff
@@ -12,10 +12,11 @@ class AuthorizingProcessor < JSONAPI::Processor
...
       set_callback :replace_fields, :before, :authorize_replace_fields
       set_callback :replace_to_one_relationship, :before, :authorize_replace_to_one_relationship
-      set_callback :create_to_many_relationship, :before, :authorize_create_to_many_relationship
-      set_callback :replace_to_many_relationship, :before, :authorize_replace_to_many_relationship
-      set_callback :remove_to_many_relationship, :before, :authorize_remove_to_many_relationship
       set_callback :remove_to_one_relationship, :before, :authorize_remove_to_one_relationship
+      set_callback :create_to_many_relationships, :before, :authorize_create_to_many_relationships
+      set_callback :replace_to_many_relationships, :before, :authorize_replace_to_many_relationships
+      set_callback :remove_to_many_relationships, :before, :authorize_remove_to_many_relationships
+      set_callback :replace_polymorphic_to_one_relationship, :before, :authorize_replace_polymorphic_to_one_relationship
```
valscion: This line is too long and will cause the linter to fail the build. Sorry about that 😥 

Maybe write it like like this?

```rb
set_callback(
  :replace_polymorphic_to_one_relationship,
  :before,
  :authorize_replace_polymorphic_to_one_relationship
)
```

handlers: :+1:

spec/dummy/app/resources/article_resource.rb
handlers: I didn't discover the reason in jsonapi-resources, but there are 22 failures if this is left out. In included_resources_spec, the failures look approximately like this: `"Param not allowed", "detail" : "id is not allowed."` I can search for more details if you like!

lib/jsonapi/authorization/authorizing_processor.rb
handlers: This makes sense to me. I will try!

handlers:
I've updated the PR to send records to the authorizer as a batch (7280547) instead of iterating over them individually. Travis is also now running successfully :). There is still an [open thread](https://github.com/venuu/jsonapi-authorization/pull/52#discussion_r106925738) related to an update required to make the dummy app's `ArticleResource` work with existing threads that we should resolve.

spec/dummy/app/resources/article_resource.rb
valscion: Hmm I wonder if it has got to with @thibaudgg's PR #11 and the changes done to the dummy app `ArticleResource` in https://github.com/venuu/jsonapi-authorization/pull/11/files#diff-cc66704bb355f2769ceb1ed76d5f02b4

I'm not really sure how the `self.verify_key` and `id=` methods work, and why exactly they were necessary to have. @thibaudgg do you have any insight to this?

lib/jsonapi/authorization/authorizing_processor.rb
valscion: Actually, I think we'd like to skip raising this here and figure out what to do with the polymorphic relationship scenario outside of this PR. This way we wouldn't add too much noise to this upgrade path.

Would you like to remove this callback from the processor?

lib/jsonapi/authorization/authorizing_processor.rb
```diff
@@ -181,27 +186,32 @@ def authorize_replace_to_many_relationship
...
-            params[:associated_key],
+          .find_by_keys(
+            params[:associated_keys],
             context: context
           )
-        related_record = related_resource._model unless related_resource.nil?
+
+        # find_by_keys doesn't raise if no records, so raise explicitly here
+        if related_resources.empty?
+          fail JSONAPI::Exceptions::RecordNotFound, params[:associated_keys]
```
valscion: Do we need this check here? What specs fail if we'd remove this part of the code?

Maybe those tests that would fail have incorrect behaviour which we'd like to change. I don't know, but this seems a bit out of place here.

lib/jsonapi/authorization/authorizing_processor.rb
handlers: Sure! Removed it here 6377da9

lib/jsonapi/authorization/authorizing_processor.rb
handlers: Good question! If this is removed, one [spec](https://github.com/handlers/jsonapi-authorization/blob/jsonapi-resources-0.9/spec/requests/relationship_operations_spec.rb#L302) failure occurs: when "authorized for remove_to_many_relationship" and "limited by policy scope on comments" it `{ is_expected.to be_not_found }`. The spec expects a 404 if the comments scope (`CommentPolicy::Scope`) is `Comment.none`.

With the new code, it instead tries to run `remove_from_comments?`, which raises `NotImplementedError`.

```ruby
  1) Relationship operations DELETE /articles/:id/relationships/comments authorized for remove_to_many_relationship limited by policy scope on comments
     Failure/Error: raise NotImplementedError

     NotImplementedError:
       NotImplementedError
     # ./spec/dummy/app/policies/article_policy.rb:44:in `remove_from_comments?'
     # ./lib/jsonapi/authorization/default_pundit_authorizer.rb:238:in `public_send'
     # ./lib/jsonapi/authorization/default_pundit_authorizer.rb:238:in `authorize_relationship_operation'
     # ./lib/jsonapi/authorization/default_pundit_authorizer.rb:183:in `remove_to_many_relationship'
     # ./lib/jsonapi/authorization/authorizing_processor.rb:201:in `authorize_remove_to_many_relationships'
     # ./spec/requests/relationship_operations_spec.rb:275:in `block (3 levels) in <top (required)>'
     # ./spec/requests/relationship_operations_spec.rb:302:in `block (5 levels) in <top (required)>'
```

I included the `fail` to preserve the old 404 behavior. Perhaps we could move that `fail` to `authorize_relationship_operation`?

valscion: Hmm I guess we'd really want to test that the `DELETE` operation authorizes properly, and doesn't use 404 when the comments can't be found.

Can you modify the tests to expect that `remove_to_many_relationship` authorizer method will be called with the proper arguments?

handlers: Sure! I have updated the tests! b4b02f5 d2e3a57

Thanks so much for the 👀!

spec/requests/relationship_operations_spec.rb
```diff
@@ -291,29 +291,33 @@
...
 
-        it { is_expected.to be_not_found }
+        let(:comments_scope) { Comment.none }
+
+        it 'calls :remove_to_many_relationship with no comments in scope' do
+          expect_operation('remove_to_many_relationship', article, [], "comments")
+          last_response
+        end
+
+        it { is_expected.not_to be_not_found }
```
valscion: Can we write this spec as either

```rb
it { is_expected.to be_successful }
```

or

```rb
it { is_expected.to be_forbidden }
```

like we have done with all the other specs?

If this feels a bit over your head, I could also tackle these at some point in the future.

spec/dummy/app/resources/article_resource.rb
thibaudgg: Hi, I'll have a look next week and give a try to this branch with our app. 
Thanks @handlers for all the work on that one! 👍

valscion: Thanks @thibaudgg! Let us know if you hit roadblocks.

These might be relevant to your tests, if you haven't updated in a while: #53 #40

handlers: Hi all! Just checking in on this — let me know if there's anything I can do to help out :)

valscion: I'm waiting on @thibaudgg to be able to test this and confirm it works nicely :)

handlers: 🆒 thanks!

thibaudgg: I launched our test suite on our CI with jsonapi-resource 0.9 and this PR and we get a bunch of errors (unexpected `406 - Not Acceptable` responses). It's hard to tell right now if it's because of upgrading to JR 0.9 or this PR specifically. 
Sadly I didn't have much time to investigate more this week. I'll try to block an hour next week.

valscion: Huh, that 406 error does indeed seem weird. It might very well be because of JR 0.9, too

spec/requests/relationship_operations_spec.rb
```diff
@@ -291,29 +291,33 @@
...
 
         it { is_expected.to be_successful }
       end
 
       context 'limited by policy scope on comments' do
-        let(:comments_scope) { Comment.none }
         before do
-          allow_operation('remove_to_many_relationship', article, kind_of(Comment), :comments)
+          allow_any_instance_of(ArticlePolicy).to receive(
+            :remove_from_comments?).and_return(true)
```
valscion: It would be great if we didn't have to worry about policies in here at all, and could create this spec in terms of only the operations stubbing instead.

Would it be possible to restructure this spec to look something like this?

```rb
context 'limited by policy scope on comments' do
  let(:comments_scope) { Comment.none }
  before do
    allow_operation('remove_to_many_relationship', article, [], :comments)
  end

  it { is_expected.to be_succesful }
end
```

This would also ensure that the `remove_to_many_relationship` is indeed called with an empty array instead of any comments.

spec/requests/relationship_operations_spec.rb
```diff
@@ -291,29 +291,33 @@
     context 'authorized for remove_to_many_relationship' do
       context 'not limited by policy scopes' do
         before do
-          allow_operations('remove_to_many_relationship', [
-            [article, comments_to_remove.first, :comments],
-            [article, comments_to_remove.second, :comments]
-          ])
+          allow_operation('remove_to_many_relationship', article, kind_of(Array), "comments")
```
valscion: Could we restructure this to test for the actual comments instead of just `kind_of(Array)`? The `kind_of(Array)` might be a bit too vague spec and could cause bugs to slip through as it doesn't care about the type of the objects inside the array itself.

Would something like this work?

```rb
allow_operation(
  'remove_to_many_relationship',
  article,
  [comments_to_remove.first, comments_to_remove.second],
  :comments
)
```

That would be the best. If the specs fail because of this due to some matchers not working correctly, we should fix the matchers instead of trying to work around them by using less specific assertions.

Also, as you can see, let's try to use the symbol version for the relationship name in the assertions (`:comments` instead of `"comments"`) or otherwise it might result in an inconsistent API with the other authorizers)

spec/requests/relationship_operations_spec.rb
```diff
@@ -282,7 +282,7 @@
 
     context 'unauthorized for remove_to_many_relationship' do
       before do
-        disallow_operation('remove_to_many_relationship', article, kind_of(Comment), :comments)
+        disallow_operation('remove_to_many_relationship', article, kind_of(Array), "comments")
```
valscion: Same thing as in https://github.com/venuu/jsonapi-authorization/pull/52/files#r109101102, just that the assertion should be `disallow_operation` instead of `allow_operation`

spec/requests/relationship_operations_spec.rb
```diff
@@ -291,29 +291,33 @@
...
+
+        it { is_expected.to be_successful }
       end
 
       # If this happens in real life, it's mostly a bug. We want to document the
       # behaviour in that case anyway, as it might be surprising.
       context 'limited by policy scope on articles' do
         before do
-          allow_operation('remove_to_many_relationship', article, kind_of(Comment), :comments)
+          allow_operation('remove_to_many_relationship', article, kind_of(Array), :comments)
```
valscion: Same as https://github.com/venuu/jsonapi-authorization/pull/52/files#r109100787

spec/support/authorization_stubs.rb
```diff
@@ -23,4 +23,11 @@ def allow_operations(operation, operation_args)
 
     allow(AUTHORIZER_CLASS).to receive(:new).with(Hash).and_return(authorizer)
   end
+
+  def expect_operation(operation, *args, authorizer: instance_double(AUTHORIZER_CLASS))
+    expect(authorizer).to receive(operation).with(*args)
+
+    allow(AUTHORIZER_CLASS).to receive(:new).with(Hash).and_return(authorizer)
+    authorizer
+  end
```
valscion: We hopefully don't need to add this new method here, if we can satisfy the spec in https://github.com/venuu/jsonapi-authorization/pull/52/files#r109100787 with the existing matchers

handlers:
Thanks so much @valscion! I am going to try to update the PR today or tomorrow.

thibaudgg:
@valscion I was able to have all our specs passed with that PR, so it looks good so far for us. :+1:

@handlers could you please rebase it on master? I would like to test it more now that #51 has been merged. Thanks!

valscion:
Thank you so much @thibaudgg for testing this out, really appreciate it!

We actually just moments ago stumbled upon an issue in JR ([it was this one if you're interested](https://github.com/cerebris/jsonapi-resources/issues/802)) and as it has been fixed in 0.9.0 we'll want to push this PR through if not today, then tomorrow.

> could you please rebase it on master?

We'd actually like merge commits to keep the history of the conversation in place, but I can do it from here on and add the test changes, too.

spec/dummy/app/resources/article_resource.rb
valscion: Ok turns out that these lines are indeed needed since v0.8.3 of JR as IDs are generated on the client side.

See https://github.com/cerebris/jsonapi-resources/releases/tag/v0.8.3

valscion:
I'll tackle this tomorrow (which in our timezone would be around 16 hours from now).

So if you feel like addressing the tests before that, @handlers, feel free to, but if you don't have the time I'll happily plow through them as well and get this PR merged ☺️ 

Thank you so much for your efforts here! 💞

handlers:
I won't be able to get to this for a while, so feel free to make any
modifications necessary!

thibaudgg:
👏

PR #53, opened 2017-03-21
Register replace_polymorphic_to_one_relationship callback

handlers:
This pull requests registers a callback on `JSONAPI::Processor#replace_polymorphic_to_one_relationship`, which is defined as an eligible callback here: https://github.com/cerebris/jsonapi-resources/blame/master/lib/jsonapi/processor.rb#L13.

I am not sure what the expected behavior ought to be, so I have simply filled in `raise NotImplementedError` so that we can decide.

This was originally part of https://github.com/venuu/jsonapi-authorization/pull/52 (comment: https://github.com/venuu/jsonapi-authorization/pull/52#discussion_r107082505). It is not necessary for tests to pass against jsonapi-resources 0.9.

### DIFF
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -16,6 +16,11 @@ module JSONAPI
       set_callback :replace_to_many_relationship, :before, :authorize_replace_to_many_relationship
       set_callback :remove_to_many_relationship, :before, :authorize_remove_to_many_relationship
       set_callback :remove_to_one_relationship, :before, :authorize_remove_to_one_relationship
+      set_callback(
+        :replace_polymorphic_to_one_relationship,
+        :before,
+        :authorize_replace_polymorphic_to_one_relationship
+      )
 
       [
         :find,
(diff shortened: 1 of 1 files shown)


handlers:
Would it be helpful for me to try to implement an authorization strategy here, rather than simply raising the `NotImplementedError`?

valscion:
> Would it be helpful for me to try to implement an authorization strategy here, rather than simply raising the NotImplementedError?

That would be lovely!

I'm not quite sure what kind of policy method we'll want to call and in what cases is the `replace_polymorphic_to_one_relationship` callback even called. Are you able to enumerate what operations to what type of resources are affected? If you're able to setup tests that expect this method to be called correctly, that would be amazing!

You might want to check out the [last comment](https://github.com/venuu/jsonapi-authorization/pull/51#issuecomment-287059591) in #51 on how authorization on resource level could also work. The #51 should also give good context on how we've thought about the relationship authorizations.

handlers:
Excellent, thanks so much. I will give it a shot!

valscion:
I'll close this PR in favor of #75. Thank you @handlers for finding this case!

PR #59, opened 2017-04-07
Add spec for missing #replace_fields scenario

gnfisher:
- Test scenarios where `replace_<type>?` is not defined for related records, not when user is authorized and unauthorized to `update?` source record.

As requested by @valscion at the end of issue #51's comments.

### DIFF
--- a/spec/jsonapi/authorization/default_pundit_authorizer_spec.rb
+++ b/spec/jsonapi/authorization/default_pundit_authorizer_spec.rb
@@ -187,7 +187,7 @@ RSpec.describe JSONAPI::Authorization::DefaultPunditAuthorizer do
 
     context 'authorized for replace_comments? and authorized for update? on source record' do
       before { stub_policy_actions(source_record, replace_comments?: true, update?: true) }
-      it { is_expected.not_to raise_error(::Pundit::NotAuthorizedError) }
+      it { is_expected.not_to raise_error }
     end
 
     context 'unauthorized for replace_comments? and authorized for update? on source record' do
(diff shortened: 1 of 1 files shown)


valscion:
Thanks for starting this PR!

When you're ready with this PR and would want a review, you can add the `Status: Ready for review` and we'll be able to more easily pick this up when we've got time :relaxed:

valscion:
These specs seem to cover a little less ground than what relationship record specs do. For example, this is done immediately inside `#replace_to_one_relationship` specs:

```rb
context 'where replace_<type>? not defined' do
  # CommentPolicy does not define #replace_article?, so #update? should determine authorization
  let(:related_records) { Article.new }
  subject(:method_call) do
    -> { authorizer.replace_to_one_relationship(source_record, related_record, :article) }
  end

  context 'authorized for update? on record' do
    before { allow_action(source_record, 'update?') }
    it { is_expected.not_to raise_error }
  end

  context 'unauthorized for update? on record' do
    before { disallow_action(source_record, 'update?') }
    it { is_expected.to raise_error(::Pundit::NotAuthorizedError) }
  end
end
```

Note that this isn't inside any

```rb
context 'unauthorized for update? on source record' do
```

blocks, so there is no extraneous stubbing going on before the specs themselves.

Do you think these specs could be made to cover a bit more ground, taking example from the relationship tests?

spec/jsonapi/authorization/default_pundit_authorizer_spec.rb
```diff
@@ -189,6 +189,10 @@
         before { stub_policy_actions(source_record, replace_comments?: false, update?: true) }
         it { is_expected.to raise_error(::Pundit::NotAuthorizedError) }
       end
+
+      context 'were replace_<type>? is undefined' do
```
valscion: Here's a bit of a typo, it probably should be "where" instead of "were"

spec/jsonapi/authorization/default_pundit_authorizer_spec.rb
```diff
@@ -208,6 +212,10 @@
         before { stub_policy_actions(source_record, replace_comments?: false, update?: false) }
         it { is_expected.to raise_error(::Pundit::NotAuthorizedError) }
       end
+
+      context 'were replace_<type>? is undefined' do
```
valscion: Here's a bit of a typo, too — it probably should be "where" instead of "were"

gnfisher:
Two issues fixed.

Issue #56, opened 2017-04-04
How to authorize relationships when creating a new resource?


valscion:
In the new resource case, we don't yet have a `record` available in the Pundit policies as all we've got is the class.

```rb
user_1 = User.new(id: 'user-1').save!
comment_1 = Comment.new(id: 'c-1', content: 'Is this the real life?').save!
comment_2 = Comment.new(id: 'c-2', content: 'Is this just fantasy?').save!
```

`POST /articles`

```json
{
  "type": "articles",
  "relationships": {
    "author": {
      "data": {
        "type": "users",
        "id": "user-1"
      }
    },
    "comments": {
      "data": [
        { "type": "comments", "id": "c-1" },
        { "type": "comments", "id": "c-2" }
      ]
    }
  }
}
```

This will need to be addressed somehow... I'm afraid that we don't actually have any methods we could use for this case, though, as all of the relationship operation methods expect that we already have a record instance! Damn.

What should we call these methods in this case? (NOTE: The names here are arbitrary, they are just meant for a conversion starter)

1. `ArticlePolicy#create?`
2. `ArticlePolicy#create_with_user?(user_1)`
3. `ArticlePolicy#create_with_comments?([comment_1, comment_2])`

This discussion was splitted off from #51

valscion:
I was thinking about authorizing the relationships in an `after` hook. That is, the order of authorizations would go like this:

`POST /articles`

1. Run `before` callback for `create_resource` operation:
  1. `JSONAPI::Authorization::AuthorizingProcessor#authorize_create_resource`
  2. `authorizer.create_resource(source_class)`
2. The new `Article` is saved (and we're still inside a transaction)
3. Run `after` callback for `create_resource` operation
  1. `JSONAPI::Authorization::AuthorizingProcessor#authorize_after_create_resource`
  2.  `authorizer.after_create_resource(source_record, related_models_with_context)`
  3. and in `after_create_resource`, call existing `#authorize_related_records` method

There's one biiig problem with this, though. That is, we would already have mutated the relations by the time the policy methods are called, so `authorize_replace_to_many_relationship` would work in a slighly different way and it would lead to weird bugs.

So we can't go the route of using after hooks.

----

We will have to come up with a way that won't have access to the newly created instance. All we've got is the record class, and the existing related models, and we will have to use only them to figure this out.

So whatever solution we choose, will have to work with an API like this:

```rb
class ArticlePolicy
  def create_with_comments?(new_comments)
[...]

PR #60, opened 2017-04-08
Authorize related records on #create_resource

gnfisher:
Based on issue #56.

- The`create_resource` method will check for a `create_with_<type>?` action on the policy. If it doesn't find one, it will pass so long as `create?` passes.

I am having a hard time getting specs in `/spec/requests/include_resources.rb` to pass. Mainly, I am struggling with the `AuthorizationStubs` class. 

The errors I can't solve all some variant of this:
```
 1) including resources alongside normal operations POST /articles/:id one-level deep has_many relationship authorized for include_has_many_resource for Comment 
     Failure/Error: authorizer.create_resource(source_class, related_models_with_context)
     
       #<InstanceDouble(JSONAPI::Authorization::DefaultPunditAuthorizer) (anonymous)> received :create_resource with unexpected arguments
         expected: (Article(id: integer, external_id: string, author_id: integer, blank_value: string), [{:relation_type=>:to_one, :relation_name=>:author, :records=>#<User id: 676971206>}, {:relation_type
..._user_id: nil>, #<Comment id: 939557610, article_id: nil, author_id: nil, reviewing_user_id: nil>]}])
              got: (Article(id: integer, external_id: string, author_id: integer, blank_value: string), [{:relation_type=>:to_one, :relation_name=>:author, :records=>#<User id: 676971206>}, {:relation_type
...nts, :records=>[#<Comment id: 939557609, article_id: nil, author_id: nil, reviewing_user_id: nil>]}])
       Diff:
       @@ -5,6 +5,5 @@
          {:relation_type=>:to_many,
           :relation_name=>:comments,
           :records=>
       -    [#<Comment id: 939557609, article_id: nil, author_id: nil, reviewing_user_id: nil>,
       -     #<Comment id: 939557610, article_id: nil, author_id: nil, reviewing_user_id: nil>]}]]
       +    [#<Comment id: 939557609, article_id: nil, author_id: nil, reviewing_user_id: nil>]}]]
       
        Please stub a default value first if message might be received with other args as well. 
     Shared Example Group: :include_directive_tests called from ./spec/requests/included_resources_spec.rb:321
     # ./lib/jsonapi/authorization/authorizing_processor.rb:115:in `authorize_create_resource'
     # ./spec/requests/included_resources_spec.rb:315:in `block (3 levels) in <top (required)>'
     # ./spec/requests/included_resources_spec.rb:45:in `block (5 levels) in <top (required)>'
```

Basically, in the spec when a new scoping is applied, the `allow_operation` method conflicts with it. This is a few levels above my usual RSpec'ing and I have banged  my head against the wall long enough not to feel too bad about asking for some help and guidance on getting this suite green again!

With help I can get this wrapped up quickly, I do have time available to get this done. Excuse the messy commits for now I will rebase and clean up when I have won my war against RSpec stubs.


gnfisher:
It occurs to me that a more ideal fallback when there is no `create_with_<type>?` action would be to check `update?` on the relationship object? That's how it was handled before.

valscion:
> It occurs to me that a more ideal fallback when there is no create_with_<type>? action would be to check update? on the relationship object? That's how it was handled before.

Yeah, I actually just answered over at #56 that this fallback logic of using `update` on the related records would make sense and would be how it was handled before :+1:

> Basically, in the spec when a new scoping is applied, the allow_operation method conflicts with it. This is a few levels above my usual RSpec'ing and I have banged my head against the wall long enough not to feel too bad about asking for some help and guidance on getting this suite green again!

Yeah those are some very tricky tests and stubbing logic. I'm not quite happy about them myself either, but I'll definitely be able to give you a hand figuring these out :)

gnfisher:
I've added the fallback to `update?` on the related record.

Whenever you have time to give me a hand with the specs I can get this wrapped up @valscion. Thanks!

valscion:
I'll likely have time this week to look into these. I really appreciate you working on these, thank you! ☺️

valscion:
Ok I managed to fix the included resources specs and pushed directly to your branch (hopefully that's OK).

I also merged in `master` here, so we're good to go with version 0.9 of JR here already :smile:

I also fixed the tricky operation specs that were failing, as now they can use this new relationship operation authorization properly! Amazing!

valscion:
> Excuse the messy commits for now I will rebase and clean up when I have won my war against RSpec stubs.

Don't worry about WIP commits, I'll squash & merge once we want to get this merged anyway so individual commits aren't that big a deal :)

valscion:
I think all we need here is the same as in the missing specs at https://github.com/venuu/jsonapi-authorization/pull/59#issuecomment-293008234

Right now we no longer have tests for the fallback scenario to `update?`

> These specs seem to cover a little less ground than what relationship record specs do. For example, this is done immediately inside `#replace_to_one_relationship` specs:
> 
> ```rb
> context 'where replace_<type>? not defined' do
>   # CommentPolicy does not define #replace_article?, so #update? should determine authorization
>   let(:source_record) { comments(:comment_1) }
>   let(:related_records) { Article.new }
>   subject(:method_call) do
>     -> { authorizer.replace_to_one_relationship(source_record, related_record, :article) }
>   end
> 
>   context 'authorized for update? on record' do
>     before { allow_action(source_record, 'update?') }
>     it { is_expected.not_to raise_error }
>   end
> 
>   context 'unauthorized for update? on record' do
>     before { disallow_action(source_record, 'update?') }
>     it { is_expected.to raise_error(::Pundit::NotAuthorizedError) }
>   end
> end
> ```
> 
> Note that this isn't inside any
> 
> ```rb
> context 'unauthorized for update? on source record' do
> ```
> 
> blocks, so there is no extraneous stubbing going on before the specs themselves.
> 
> Do you think these specs could be made to cover a bit more ground, taking example from the relationship tests?

gnfisher:
Awesome, thank you. I will take a look today and see if I can't get all the specs done on this and on #59 today and get this wrapped up!

gnfisher:
I think this covers the missing scenarios. Please let me know if more is needed.

spec/jsonapi/authorization/default_pundit_authorizer_spec.rb
```diff
@@ -213,55 +213,66 @@
...
+        before do
+          stub_policy_actions(source_class, create?: true)
+          related_records.each { |r| stub_policy_actions(r, update?: true) }
+        end
+        it { is_expected.not_to raise_error(::Pundit::NotAuthorizedError) }
       end
 
-      context 'unauthorized for update? on any of the related records' do
-        let(:related_records) { [Comment.new(id: 1), Comment.new(id: 2)] }
+      context 'unauthorized for update? on related records' do
```
valscion: This might be a bit of a nitpick, but do you think it'd be a good idea to test this as

```
context 'unauthorized for update? on any of the related records'
```

...and then just stub one of the `related_records` to return `false` for `update?` and stub others to return `true`? It would be similar to what there used to be.

gnfisher: Yes, I will change it. It's a better test.

spec/jsonapi/authorization/default_pundit_authorizer_spec.rb
```diff
@@ -213,55 +213,66 @@
...
-      context 'authorized for update? on all of the related records' do
-        before { related_records.each { |r| allow_action(r, 'update?') } }
-        it { is_expected.to raise_error(::Pundit::NotAuthorizedError) }
+    context 'authorized for create? where create_with_<type>? is undefined' do
+      context 'authorized for update? on related records' do
+        before do
+          stub_policy_actions(source_class, create?: true)
+          related_records.each { |r| stub_policy_actions(r, update?: true) }
+        end
+        it { is_expected.not_to raise_error(::Pundit::NotAuthorizedError) }
```
gnfisher: Yeah, I should have known better on this one as just a few weeks back I was reading about this same sort of scenario. Will update!

PR #65, opened 2017-05-21
Handle has_one properly

nbarthel:
Fix issue with has_one where iterator is applied to object causing a 500.

Spacing is goofy, I blame Atom.

### DIFF
--- a/lib/jsonapi/authorization/default_pundit_authorizer.rb
+++ b/lib/jsonapi/authorization/default_pundit_authorizer.rb
@@ -115,6 +115,7 @@ module JSONAPI
                     policy: policy
             end
           else
+	    records = records.is_a?(Array) ? records : [records]
             records.each do |record|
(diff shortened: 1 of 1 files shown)


kerrizor:
A cleaner way might be to take advantage of `Kernel.Array`

```
2.4.0 :001 > arr = ["foo"]
 => ["foo"]
2.4.0 :002 > Array(arr)
 => ["foo"]
2.4.0 :003 > Array("bar")
 => ["bar"]
```

This would let you change the iterator line to simply

```
Array(records).each do |record|
```

valscion:
Thanks for noticing this case, @nbarthel! Seems like our test suite was missing this scenario, so the changes I want are a bit larger than this so I opened a PR to fix this bug. See #66

valscion:
Thanks for opening this PR, a fix for the underlying cause is now part of `v1.0.0.alpha3` ☺️

PR #66, opened 2017-05-22
Fix create_resource authorization breaking with has_one association

valscion:
We seem to have missed a test case for the `related_records_with_context` cases and so have got this bug. Credits to @nbarthel for discovering this case in #65, thanks!

Closes #65

### DIFF
--- a/lib/jsonapi/authorization/default_pundit_authorizer.rb
+++ b/lib/jsonapi/authorization/default_pundit_authorizer.rb
@@ -115,7 +115,7 @@ module JSONAPI
                     policy: policy
             end
           else
-            records.each do |record|
+            Array(records).each do |record|
               ::Pundit.authorize(user, record, 'update?')
             end
           end
--- a/spec/dummy/app/policies/article_policy.rb
+++ b/spec/dummy/app/policies/article_policy.rb
@@ -32,6 +32,10 @@ class ArticlePolicy
     raise NotImplementedError
   end
 
+  def create_with_author?(_author)
+    raise NotImplementedError
+  end
+
   def create_with_comments?(_comments)
     raise NotImplementedError
   end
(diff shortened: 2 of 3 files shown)


valscion:
Do these changes look good to you, @gnfisher? After all, you were the one these relationship checks in #60 so I value your input on this.

Do you think the `data[:records]` would need a rename in the `related_records_with_context` as it can contain just a single record, too? I don't yet have a good idea on what to call it, if we deem `records` can be a bit deceiving. Do you have any ideas?

gnfisher:
The changes look good. I'm trying to remember back to this and how the `has_one` scenario wasn't an issue that came up in my use/testing (we're using a version of this in our project at work and haven't run up against this scenario - thankfully ;)). 

As for naming the variable, this morning nothing strikes me as any better than `records` off hand. Ideally I'd like for `data[:records]` to always be an Array, which I'm sure is what I thought I was doing at the time. This way, even if it's a singular `has_one` relationship, it is still an array, just one with a single item.

valscion:
> Ideally I'd like for data[:records] to always be an Array, which I'm sure is what I thought I was doing at the time. This way, even if it's a singular has_one relationship, it is still an array, just one with a single item.

Yeah I thought about this as well, but then it would make some quite funky code in `authorize_related_resources` when [`replace_to_one_relationship` is called](https://github.com/venuu/jsonapi-authorization/blob/v1.0.0.alpha2/lib/jsonapi/authorization/default_pundit_authorizer.rb#L250). We'd have to unwrap the array and only take the first item from there, in order to not to change the API of `replace_to_one_relationship` to also accept an array of records. 😕

valscion:
Oh, and if the changes look good to you, could you approve the review? The big red "Review required" text doesn't really encourage me to bypass it :smile:

screen shot 2017-05-22 at 16 14 17

PR #69, opened 2017-05-29
Do not throw not found error when resource relationship is null

jpalumickas:
This PR extends https://github.com/venuu/jsonapi-authorization/pull/54 with added tests.


valscion:
Ok, now I see what's happening. Thanks for the PR!

Do you think it would be better if it were somehow possible to tell the authorizer that the relationship is about to get emptied? What if the expected authorizer call would look like this instead?

```rb
allow_operation(
  'replace_fields',
  article,
  [{
    relation_name: :author,
    relation_type: :to_one,
    records: nil
  }]
)
```

These kind of relationship specs currently reside in a slightly badly named file `spec/requests/tricky_operations_spec.rb`. Do you think you could move the added spec there, and maybe structure it like so?

```rb
describe 'PATCH /articles/:id (nullifying to-one relationship)' do
  # spec code that is similar to the
  # 'PATCH /articles/:id (mass-modifying relationships)' spec code
  # ...
end
```

Would you also be open to change the behavior to match that?

Thanks for your contribution! If you want me to open up something more, feel free to ask me anything :relaxed:

jpalumickas:
@valscion 
Moved tests to tricky operations.



> Do you think it would be better if it were somehow possible to tell the authorizer that the relationship is about to get emptied? What if the expected authorizer call would look like this instead?

For this one I had error, so I left as it is for now:
```rb
#<InstanceDouble(JSONAPI::Authorization::DefaultPunditAuthorizer) (anonymous)> received :replace_fields with unexpected arguments
         expected: (#<Article id: 1012824740, external_id: "3", author_id: 676971205, blank_value: nil>, [{:relation_name=>:author, :relation_type=>:to_one, :records=>nil}])
              got: (#<Article id: 1012824740, external_id: "3", author_id: 676971205, blank_value: nil>, [])
```

valscion:
> For this one I had error, so I left as it is for now:

Ah, but actually that error highlights the behavior that we would actually want to have. We would want to allow the authorizer to have a way of handling this situation.

However, in the [`DefaultPunditAuthorizer`](https://github.com/venuu/jsonapi-authorization/blob/v1.0.0.alpha3/lib/jsonapi/authorization/default_pundit_authorizer.rb#L262-L264), we might want to actually call `remove_to_one_relationship` when the `records` is `nil` so that the policy method that would end up being called would be `"remove_#{relationship_type}?"`.

Do you want to go ahead and do these changes? That would mean these changes:

- [ ] Change the [`next` in `AuthorizingProcessor#related_models_with_context`](https://github.com/jpalumickas/jsonapi-authorization/blob/5f3f37282710419bc15a23d9602fb6e90f6cee1d/lib/jsonapi/authorization/authorizing_processor.rb#L275) to `nil` so that the authorizer class gets `nil` in the `records` part of the hash
  - [ ] This would mean that the test error you saw and didn't push would pass, so the spec could be updated
  - [ ] You could then also removed the `.compact` in the end of the `#related_models_with_context` method as it is no longer necessary
- [ ] Update [the part in`DefaultPunditAuthorizer#authorize_related_resources` where it handles `:to_one` relation type](https://github.com/venuu/jsonapi-authorization/blob/v1.0.0.alpha3/lib/jsonapi/authorization/default_pundit_authorizer.rb#L262-L264) to check whether `records` is `nil` and if so, call [`remove_to_one_relationship`](https://github.com/venuu/jsonapi-authorization/blob/v1.0.0.alpha3/lib/jsonapi/authorization/default_pundit_authorizer.rb#L199)
  - [ ] Add a third `describe` block inside the `DefaultPunditAuthorizer#replace_fields` specs that is almost 1-to-1 the same as the [`with "relation_type: :to_one"`](https://github.com/venuu/jsonapi-authorization/blob/v1.0.0.alpha3/spec/jsonapi/authorization/default_pundit_authorizer_spec.rb#L163-L208) block but just with the description of `with "relation_type: :to_one" and records is nil`

Does this sound reasonable to you?

lib/jsonapi/authorization/authorizing_processor.rb
```diff
@@ -249,6 +249,8 @@ def related_models
         [:to_one, :to_many].flat_map do |rel_type|
           data[rel_type].flat_map do |assoc_name, assoc_value|
             case assoc_value
+            when nil
+              next
```
valscion: I don't think we need to change the `def related_models` method at all. Your bug fix should only be necessary to touch the `related_models_with_context` method.

spec/requests/tricky_operations_spec.rb
```diff
@@ -168,4 +168,30 @@
...
+      }
+      EOS
+    end
+    let(:policy_scope) { Article.all }
+    let(:user_policy_scope) { User.all }
+    subject(:last_response) { patch("/articles/#{article.external_id}", json) }
+
+    before do
+      allow_any_instance_of(UserPolicy::Scope).to receive(:resolve)
+        .and_return(user_policy_scope)
```
valscion: This spec would actually pass even if we didn't handle `UserPolicy::Scope` at all, so I guess these `user_policy_scope` things could be removed.

jpalumickas:
@valscion Fixed 😉

valscion:
Wow, looking great! Thanks for the fast turn-around 💞 

Are you OK with me adding you to the list of contributors in the README?

jpalumickas:
Sure, cheers ! 😉

valscion:
I'll release alpha4 version after I've had lunch. Thanks for your contributions!

jpalumickas:
Thanks!

valscion:
Released v1.0.0.alpha4 with this change :)

jpalumickas:
Perfect! thanks for the release 🚀

PR #71, opened 2017-06-04
Add thorough documentation about relationship operations

valscion:
The goal of this PR is to make it more clear how this gem works and why.

I want to make it clear what relationship policy methods we will call when relationships change and what methods the fallback ends up calling.

## For reviewers:

Start reading docs from [README.md](https://github.com/venuu/jsonapi-authorization/blob/docs/README.md) and then follow through to the [relationship authorization documentation](https://github.com/venuu/jsonapi-authorization/blob/docs/docs/relationship-authorization.md)


docs/relationship-authorization.md
```diff
@@ -0,0 +1,384 @@
+# <a name="doc-top"></a>Authorization of operations touching relationships
+
+`JSONAPI::Authorization` (JA) is unique in the way it considers relationship changes to change the underlying models. Whenever an incoming requests changes associated resources, JA will authorize those operations are OK.
```
olleolleolle: Typo: `an incoming requests` => `an incoming request`

valscion: Thanks!

valscion:
> I read all of it, and found only a single typo.

Thanks! What do you think of the structure — would there be a better way to do this?

olleolleolle:
@valscion I am a sucker for narrative documentation, and a longer example always wins me over.

I know that this is the Hard Part and Too Much and so on, but now I have said it.

Does Pundit have any such long example you could extend?

valscion:
> I know that this is the Hard Part and Too Much and so on, but now I have said it.

It's OK, I am looking for feedback precisely like this ☺️. I am used to critique and am indeed looking for negative things in here as well.

> Does Pundit have any such long example you could extend?

I didn't find long examples from https://github.com/elabs/pundit/ — if you find some good example, then I'd be happy to check them out :)

> I am a sucker for narrative documentation, and a longer example always wins me over.

Yeah I also like long examples. I tried to make these docs have as much examples as possible, but also cover the entire API surface... maybe it's too much to do both at once?

On what level would you like to see the examples? Would it need to be a proper application that I could showcase to display how this works? That is, would the examples indeed cover a "real-world" use-case instead of this minimal example setup that here currently is?

> The examples on this page use these models:
> 
> ```rb
> class Article < ActiveRecord::Base
>   has_many :comments
>   belongs_to :author, class_name: 'User'
> end
> 
> class Comment < ActiveRecord::Base
>   belongs_to :article
> end
> 
> class User < ActiveRecord::Base
>   has_many :articles, foreign_key: :author_id
> end
> ```
> 
> ...and these resources:
> 
> ```rb
> class ArticleResource < JSONAPI::Resource
>   include JSONAPI::Authorization::PunditScopedResource
> 
>   # `acts_as_set` allows replacing all comments at once
>   has_many :comments, acts_as_set: true
>   has_one :author, class_name: 'User'
> end
> 
> class CommentResource < JSONAPI::Resource
>   include JSONAPI::Authorization::PunditScopedResource
> 
>   has_one :article
> end
> 
> 
> class UserResource < JSONAPI::Resource
>   include JSONAPI::Authorization::PunditScopedResource
> 
>   has_many :comments
> end
> ```

olleolleolle:
@valscion Blog posts and authors and comments - that IS the canonical REST example. Let's stay with that, since it's the Hello World of this domain.

valscion:
Yeah, this should be enough to cover the current API surface.

I'm not sure whether it's a good idea to have all of these on a single page, though, as it gets quite long 🤔 . Also, I was wondering if it would be possible to have a condensed table-like summary of all the custom policy methods used and their fallbacks.

olleolleolle:
@valscion If you're excited about integrated documentation in the code-base, you could extend YARD. 

Once, I added a customization of YARD to a project to render an HTML table of config options.

https://github.com/gocardless/hutch/blob/master/.yardopts links to the thing. You can poke around there to see how that is done.

valscion:
Hmm, where can I see from there the output of YARD? Is it the README? Or where?

olleolleolle:
@valscion Yes, the generated content is cut-and-pasted into the README.

olleolleolle:
@valscion The _un-configured_ YARD output can be found in the rubydoc website: http://www.rubydoc.info/gems/hutch

valscion:
Hmm, we do have rubydoc as well: http://www.rubydoc.info/github/venuu/jsonapi-authorization

We could add more examples among the API docs for sure, if that's what you mean?

valscion:
Especially the docs for [`DefaultPunditAuthorizer`](http://www.rubydoc.info/github/venuu/jsonapi-authorization/JSONAPI/Authorization/DefaultPunditAuthorizer) are ones we've put some effort into.

valscion:
> Thank you for filling in gaps and improving the accuracy of this document!

Yeah this will soon be complete, I'm just writing the last bits :)

valscion:
If anyone from @venuu/core could check this documentation out and ask stupid questions, it would be awesome :). Same goes for @matthias-g and @gnfisher as you two have been a tremendous help in making this a reality.

valscion:
Thanks @jraunio! In that case, I'm gonna merge this :relaxed:

PR #75, opened 2017-07-05
Authorize replacing of a polymorphic has-one relationship

valscion:
This PR picks off from #53 and adds authorization for the replacing of a polymorphic has-one relationship.

Seems like `jsonapi-resources` is broken and won't allow clearing of a polymorhpic has-one relationship by sending

`PATCH /tags/1/relationships/taggable`

```
{ "data": null }
```

The issue upstream is here: https://github.com/cerebris/jsonapi-resources/issues/1081

I skipped the tests for the relationship nullification for now.

### DIFF
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -222,7 +222,34 @@ module JSONAPI
       end
 
       def authorize_replace_polymorphic_to_one_relationship
-        raise NotImplementedError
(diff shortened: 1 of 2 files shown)


valscion:
Ach this won't work if the associated polymorphic class changes. Or I'm not sure if it works... will ave to check that.

valscion:
Ok I think that's it, now it should work even if the type of the associated object changes with the replace operation.

PR #81, opened 2017-08-31
Fix relation_name lambda being broken

### LINKED ISSUE #80: JR lamda relation_name gets broken
The ability to use context-driven relation names seems to be broken, because the context argument isn't passed as required to `JSONAPI::Relationship.relation_name` when invoked from within jsonapi-authorization.

> relation_name - the name of the relation to use on the model. A lambda may be provided which allows conditional selection of the relation based on the context.

When JR invokes the lambda, the received argument looks something like this:
`{:context=>{:current_user=>#<User id: 14, ... ...>, :demo_mode=>true, :action=>"show"}, :include_directives=>#<JSONAPI::IncludeDirectives:0x007f6b5d043290 @resource_klass=API::V1::SomeResource, @force_eager_load=false, @include_directives_hash={:include_related=>{:some_included=>{:include=>true, :include_related=>{}, :include_in_join=>true}}}>, :fields=>{}}`

But when invoked from `JSONAPI::Authorization::AuthorizingProcessor.authorize_include_item` at `authorize_include_item(resource_klass, source_record, rel_name)`, rather than passing the full context object, the lambda receives:
`{:current_user=>#<Tibbing::User id: 14, ...>, :demo_mode=>true, :action=>"show"}`

Which, since the context is at least present, could be cludged around, but unfortunately, it is then invoked from `JSONAPI::Authorization::PunditScopedResource.fetch_relationship` via `v.relation_name({})` so we get `{}`, there is no way for the lamda to return the correct relation_name, and it errors out, disabling this functionality.

In our context, the `demo_mode` context boolean needs to 'pick' the appropriate result of a to_one relationship.  I'm going to have another crack at over-riding records_for_ in the resource, but it would be good to ensure that the JR context is always provided to .relation_name, and test that lambda relation_names work 'as advertised' by JR.

I just wish I had the time to work on tweaking the gem and send through a pull-request, rather than just adding problems ... :-(

valscion:
This fixes #80

It seems to be that JR always passes a hash to the `relation_name` accessors, with the `context` under key `:context`

https://github.com/cerebris/jsonapi-resources/blob/v0.9.0/lib/jsonapi/resource.rb#L250

Let's just use that everywhere in `jsonapi-authorization`, too.

It's a bit odd that JR passes way more information down the same `relation_name` when in the scope of processor, but that doesn't seem to be important to us. Documentation only refers to fetching the `context` from the options, and AFAICT, doesn't tell about what else is in the options hash:

http://jsonapi-resources.com/v0.9/guide/resources.html#Options

```rb
class BookResource < JSONAPI::Resource
  # Only book_admins may see unapproved comments for a book. Using
  # a lambda to select the correct relation on the model
  has_many :book_comments, relation_name: -> (options = {}) {
    context = options[:context]
    current_user = context ? context[:current_user] : nil
    unless current_user && current_user.book_admin
      :approved_book_comments
    else
      :book_comments
    end
  }
  ...
end
```

### DIFF
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -347,7 +347,7 @@ module JSONAPI
             next_resource_klass = relationship.resource_klass
             Array.wrap(
               source_record.public_send(
-                relationship.relation_name(context)
+                relationship.relation_name(context: context)
               )
             ).each do |next_source_record|
               deep.each do |next_include_item|
@@ -364,7 +364,7 @@ module JSONAPI
           case relationship
           when JSONAPI::Relationship::ToOne
             related_record = source_record.public_send(
-              relationship.relation_name(context)
+              relationship.relation_name(context: context)
             )
             return if related_record.nil?
             authorizer.include_has_one_resource(source_record, related_record)
--- a/lib/jsonapi/authorization/pundit_scoped_resource.rb
+++ b/lib/jsonapi/authorization/pundit_scoped_resource.rb
@@ -31,7 +31,7 @@ module JSONAPI
 
       def fetch_relationship(association_name)
         relationships = self.class._relationships.select do |_k, v|
-          v.relation_name(context) == association_name
+          v.relation_name(context: context) == association_name
         end
         if relationships.empty?
           nil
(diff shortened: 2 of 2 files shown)


valscion:
@Justin-Maxwell, could you test this out to make sure that this indeed fixes your problems found in #80? You should be able to install this version by listing `jsonapi-authorization` like this in your `Gemfile`:

```
gem 'jsonapi-authorization', github: 'venuu/jsonapi-authorization', branch: 'fix-context-passing'
```

valscion:
Hey @Justin-Maxwell, any chance to test this branch out to see if it fixes your issue?

valscion:
I'll just assume that this fixed #80 — this change should make sense either way.

Issue #64, opened 2017-04-15
Compatibility with jsonapi-resources 0.10


valscion:
Seems like we will have to do some work in the future to be compatible with upcoming JR version, 0.10, where resources being closely coupled to ActiveRecord is being worked on.

@hidde-jan in https://github.com/venuu/jsonapi-authorization/issues/48#issuecomment-293917276

> Are there any thoughts on how (or if) to incorporate https://github.com/cerebris/jsonapi-resources/issues/1006 and https://github.com/cerebris/jsonapi-resources/pull/977?
> 
> From JA:R 0.10 onward, the idea is to delegate record fetching to a `RecordAccessor`.
> 
> In the future, the logic in PunditScopedResource might need to be moved into such a class.

---

EDIT:
* See #82 for a case where a new upcoming release caused compatibility issues with upstream

valscion:
Ok JR 0.10 has been released now. It seems that the major changes done are things concerning jsonapi-authorization, so it will take quite an effort to get us compatible with it.

https://jsonapi-resources.com/v0.10/guide/whats_new.html

We should update this gem to be compatible with JR 0.10 and release a new major version to support only that version. People on older versions could use older versions of jsonapi-authorization.

Any help is appreciated. I don't have time any time soon to look into this and neither is this a priority on our team any time soon.

rpbaltazar:
@valscion do you have any sort of tasklist of things that need to be touched ?

valscion:
Not really, sorry.

One thing I'd really like to see here would be a new approach to tests where we would assert on user visible functionality — so going from a full request to an expected authorization error without relying on any stubbing.

Some of the current tests have gotten so complex due to the authorization stubbing going on that I have a lack of trust on any big changes done to the existing tests.

These new types of tests could first start small and test only a subset of the authorization logic. I'd like them to be in a totally different directory than the existing RSpec tests, and as self-contained as possible. Meaning that the resources, controllers and policies needed for the tests would all be in one file, even if it would make for a longer file.

This type of tests could be done with minitest instead of RSpec, too.

romikoops:
Hello, any progress with supporting 0.10? 4 months since issue opening.

valscion:
Does not seem like it. Feel free to offer help if you need JR 0.10 support.

lgebhardt:
[...]

PR #82, opened 2017-09-07
Fix a bug for checking JSONAPI::Relationship::ToOne authorization

clemthenem:
I had an error when I wanted to fetch a relationship and managed to fix it like this.
I'm not sure if that is the most correct way to do it but it worked for me.
Thanks for your efforts!

### DIFF
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -66,21 +66,20 @@ module JSONAPI
           params[:parent_key],
           context: context
         )
+        parent_record = parent_resource._model
 
         relationship = @resource_klass._relationship(params[:relationship_type].to_sym)
 
-        related_resource =
+        related_record =
           case relationship
           when JSONAPI::Relationship::ToOne
-            parent_resource.public_send(params[:relationship_type].to_sym)
+            parent_record.public_send(params[:relationship_type].to_sym)
           when JSONAPI::Relationship::ToMany
             # Do nothing — already covered by policy scopes
           else
             raise "Unexpected relationship type: #{relationship.inspect}"
           end
 
-        parent_record = parent_resource._model
-        related_record = related_resource._model unless related_resource.nil?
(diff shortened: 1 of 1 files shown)


valscion:
Hi! Thanks for your PR :)

> I had an error

Could you clarify a bit? I'd need to know what the exact error was so I could figure out what this code is fixing. Ideally, I'd like to get a spec case to avoid a similar error from happening in the future, so knowing what error this fixes is crucial :smile:

clemthenem:
Yes sure! 
So let's say we have to JSONAPI::Resources like this:
`
module Api::V1
  class UserResource < BaseResource
    has_one :profile
  end
end


module Api::V1
  class ProfileResource < BaseResource
    has_one :user
  end
end
`

If I try to call the following endpoint:
`localhost:3000/api/v1/profiles/XXXXXXXXX/relationships/user`
or this one
`localhost:3000/api/v1/profiles/XXXXXXXXX/user`

I get the following exception:
`"exception": "undefined method `user' for #<Api::V1::ProfileResource:0x00000006265ec0>",
                "backtrace": [
                    "/home/clement/.rvm/gems/ruby-2.4.1/bundler/gems/jsonapi-authorization-24bfed7949a3/lib/jsonapi/authorization/authorizing_processor.rb:75:in `public_send'",`

I hope my example makes sense? Do not hesitate if you need me to clarify

valscion:
Hmm, can you tell me what version of `jsonapi-resources` you are using? And what version of `jsonapi-authorization`?

clemthenem:
I'm on both master branches

valscion:
OK, that makes sense. We do not yet support the upcoming `jsonapi-resources` v0.10 (see #64), so could you try downgrading `jsonapi-resources` to 0.9.0 and see if your error goes away?

clemthenem:
Damn, sorry about that... Yes the errors goes away in the previous version!

valscion:
Yess, great to hear that! I'm gonna close this issue, and add a link to #64 to notice that there will be trouble heading down this road :)

Issue #48, opened 2017-02-20
Roadmap for version 1.0 🎉


valscion:
This is a roadmap for getting to version 1.0.

- [x] [#40] Allow authorization of relationship operations separately
- [x] [#47] Add exhaustive tests for the new relationship operation policies
- [x] [#95] The API of `DefaultPunditAuthorizer` methods could be changed to use keyword arguments
  * Currently the API change is backwards-incompatible in a way that could be very confusing,
    namely as `replace_to_one_relationship` has the same amount of parameters but are very
    different than what there was before
  * Keyword arguments are supported ever since Ruby 2.0 so we can use them
- [x] [#51] Look into the API of `DefaultPunditAuthorizer#replace_fields` and figure out if it could be implemented similarly to #40 did
  * The replace operation could even call the new replace/remove relationship authorization methods
  * If this wouldn't be done, we might need to bump to v2.0.0 after implementing as it would be a major breaking change in the API... so better look into it sooner than later
- [x] [#60] Authorize relationships when creating a new resource
- [**PR / Help wanted**] Document a design pattern for specifying relationship operations logic only once
  * https://github.com/venuu/jsonapi-authorization/issues/30#issuecomment-280854269 is a good first start
  * **EDIT on 2019-01-22:** After a lack of contributions, this was postponed to a future time.
- [x] [#52] Fix compatibility with [v0.9.0](https://github.com/cerebris/jsonapi-resources/releases/tag/v0.9.0) of `jsonapi-resources` #36
- [x] [#75] Figure out what to do with the `replace_polymorphic_to_one_relationship`
  * This case that was found in https://github.com/venuu/jsonapi-authorization/pull/52#discussion_r106925265
- [x] [#98] Test with Ruby 2.3
- [x] [#98] Test with rails 4.2, 5.0, 5.1 and 5.2
- [x] Release an alpha of 1.0.0 and gather feedback
- [x] Start using the latest release internally at Venuu
- [x] Release a beta of 1.0.0 and gather feedback

...is there anything I've missed?

This roadmap is intended to be a live documentation that can be changed later as need be.

valscion:
I just published v1.0.0.alpha1 with JR 0.9 compatibility and finer grained authorization for relationship operations and `replace_fields` scenario. Feel free to take that for a spin and report back issues you find :relaxed:

/cc @thibaudgg, you might be interested in using the alpha release if you want to use JR 0.9

thibaudgg:
[...]

Issue #77, opened 2017-08-20
Upgrade from alpha2 to alpha5 causes "unable to find policy `HashPolicy`"


arcreative:
I have a feeling this is because I'm using my own authorizer, but it looks like https://github.com/venuu/jsonapi-authorization/commit/7ca89d50c5a17d77245312fd4bbc0b1d7e8ab1ff breaks your old suggestion to override changes to relationships as changes to the record itself:

`custom_pundit_authorizer.rb`
```rb
require 'pundit'

module JSONAPI
  module Authorization
    class CustomPunditAuthorizer < DefaultPunditAuthorizer

      def replace_fields(source_record, new_related_records)
        ::Pundit.authorize(user, source_record, 'update?')

        new_related_records.each do |record|
          ::Pundit.authorize(user, record, 'show?')
        end
      end

      def create_resource(source_class, related_records)
        ::Pundit.authorize(user, source_class, 'create?')

        related_records.each do |record|
          ::Pundit.authorize(user, record, 'show?')
        end
      end
    end
  end
end
```

That said, was there something in the works to relax these constraints with an initializer config option?

Full trace:
```sh
Internal Server Error: unable to find policy `HashPolicy` for `{:relation_type=>:to_one, :relation_name=>:bundle, :records=>#<Bundle id: "123", name: "Yadda yadda", created_at: "2017-07-03 17:53:13", updated_at: "2017-07-03 17:53:13">}` /Users/adam/.rvm/gems/ruby-2.3.1/gems/pundit-1.1.0/lib/pundit/policy_finder.rb:59:in `policy!'
/Users/adam/.rvm/gems/ruby-2.3.1/gems/pundit-1.1.0/lib/pundit.rb:112:in `policy!'
/Users/adam/.rvm/gems/ruby-2.3.1/gems/pundit-1.1.0/lib/pundit.rb:62:in `authorize'
/Users/adam/repos/arcreative/my-project/lib/jsonapi/authorization/custom_pundit_authorizer.rb:19:in `block in create_resource'
/Users/adam/repos/arcreative/my-project/lib/jsonapi/authorization/custom_pundit_authorizer.rb:18:in `each'
/Users/adam/repos/arcreative/my-project/lib/jsonapi/authorization/custom_pundit_authorizer.rb:18:in `create_resource'
[...]

Issue #39, opened 2016-11-11
Authorizing create/update by context of the instance of the record class being updated?


carlthuringer:
Hi, just digging into using Pundit and jsonapi-authorization on a fairly large project and ran into an issue with authorizing the create. 

My `create?` method receives the Class, not the instance being created:
`ChainSetting(id: integer, chain_id: integer, key: string, value: string, created_at: datetime, updated_at: datetime)`

In this case, I have authorized a specific user account and within the tree of my models, I can traverse from that specific user down to a specific record of a chain, validating ownership of the chain.

So what I want to do is, given the user has made a create request to create a new `ChainSetting` with a `chain_id: 1234`, I want to lookup all the user's chains (I can do this, just join some rows) and validate that `1234` is in fact one of the user's chains.

But instead the model passed to the Policy is just the raw Class, not an instance! Awwugh!

I took a quick look through the source code and it seems the Processor is responsible for deciding what to send to the authorization policies. I'm about to check out for the day so I can't dive any deeper into it.

It seems to me that there's a flaw here. If API users are authorized to blindly create, then they can create relationships for each others' records, and we don't want that at all!

You may be noticing that I'm setting a foreign key like it's an attribute. Yes, that's strange, but not disallowed in json api. We have a convoluted legacy database here with foreign key constraints and hooks and callbacks galore that make some more orthodox updates difficult.

Regardless of this breach of protocol, one should be able to say that a `Red User` is authorized to update "ChainSetting" with `door_color: red` but not `door_color: black`, if that's what we want to do. In normal Pundit you can pass an instance and validate just like that, but in jsonapi-authorization it's not possible.

What are your thoughts? Shouldn't the instance being authorized be passed to the Policy?

valscion:
You can check out `DefaultPunditAuthorizer` class to see how each operation is mapped to a Pundit policy check. As you found out, `create` gets the class and not the instance itself. However, we also restrict you from creating relationships to objects you don't have an `update` permission for, [see the code](https://github.com/venuu/jsonapi-authorization/blob/v0.8.1/lib/jsonapi/authorization/default_pundit_authorizer.rb#L102-L116).

We authorize actions at the edge of the API before any models are changed or objects are built. If you're looking for authorization on a `before_save` hook, you might find that [`jsonapi-resources`](https://github.com/cerebris/jsonapi-resources) suits you better.

[...]

PR #95, opened 2018-06-28
Default pundit authorizer keyword args

nruth:
I saw in #48 that help was wanted using keyword args for `JSONAPI::Authorization::DefaultPunditAuthorizer` so I've had a go at it. I've done 2 methods to see how it goes, and I'd like feedback to check I'm doing what you wanted before continuing. 

The only hard part I found was making the `AuthorizationStubs` helpers work with keyword args as well as splat args. I'm not entirely happy with how I solved that, so any thoughts on that would be great. To see why I've done that, try changing the stub method and re-run the tests and you'll find a lot of things start breaking in various places with wrong-number-of-argument errors and other fun things.

### DIFF
--- a/spec/support/authorization_stubs.rb
+++ b/spec/support/authorization_stubs.rb
@@ -2,22 +2,18 @@ module AuthorizationStubs
   AUTHORIZER_CLASS = JSONAPI::Authorization::DefaultPunditAuthorizer
 
   def allow_operation(operation, *args, authorizer: instance_double(AUTHORIZER_CLASS), **kwargs)
-    if kwargs.empty?
-      allow(authorizer).to receive(operation).with(*args).and_return(nil)
-    else
-      allow(authorizer).to receive(operation).with(*args, **kwargs).and_return(nil)
-    end
+    allow(authorizer).to receive(operation).tap {|x|
+      if kwargs.empty? then x.with(*args) else x.with(*args, **kwargs) end
+    }.and_return(nil)
 
     allow(AUTHORIZER_CLASS).to receive(:new).with(Hash).and_return(authorizer)
     authorizer
   end
 
   def disallow_operation(operation, *args, authorizer: instance_double(AUTHORIZER_CLASS), **kwargs)
-    if kwargs.empty?
-      allow(authorizer).to receive(operation).with(*args).and_raise(Pundit::NotAuthorizedError)
-    else
-      allow(authorizer).to receive(operation).with(*args, **kwargs).and_raise(Pundit::NotAuthorizedError)
-    end
+    allow(authorizer).to receive(operation).tap {|x|
+      if kwargs.empty? then x.with(*args) else x.with(*args, **kwargs) end
+    }.and_raise(Pundit::NotAuthorizedError)
 
     allow(AUTHORIZER_CLASS).to receive(:new).with(Hash).and_return(authorizer)
     authorizer
--- a/spec/support/authorization_stubs.rb
+++ b/spec/support/authorization_stubs.rb
@@ -1,15 +1,23 @@
 module AuthorizationStubs
   AUTHORIZER_CLASS = JSONAPI::Authorization::DefaultPunditAuthorizer
 
-  def allow_operation(operation, *args, authorizer: instance_double(AUTHORIZER_CLASS))
-    allow(authorizer).to receive(operation).with(*args).and_return(nil)
+  def allow_operation(operation, *args, authorizer: instance_double(AUTHORIZER_CLASS), **kwargs)
+    if kwargs.empty?
+      allow(authorizer).to receive(operation).with(*args).and_return(nil)
+    else
+      allow(authorizer).to receive(operation).with(*args, **kwargs).and_return(nil)
+    end
 
     allow(AUTHORIZER_CLASS).to receive(:new).with(Hash).and_return(authorizer)
     authorizer
   end
 
-  def disallow_operation(operation, *args, authorizer: instance_double(AUTHORIZER_CLASS))
-    allow(authorizer).to receive(operation).with(*args).and_raise(Pundit::NotAuthorizedError)
+  def disallow_operation(operation, *args, authorizer: instance_double(AUTHORIZER_CLASS), **kwargs)
+    if kwargs.empty?
+      allow(authorizer).to receive(operation).with(*args).and_raise(Pundit::NotAuthorizedError)
+    else
+      allow(authorizer).to receive(operation).with(*args, **kwargs).and_raise(Pundit::NotAuthorizedError)
+    end
 
     allow(AUTHORIZER_CLASS).to receive(:new).with(Hash).and_return(authorizer)
     authorizer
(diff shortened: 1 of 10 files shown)


valscion:
Hi! Thanks for taking a stab at this :relaxed:

Looks like the tests are passing right now, nice :+1:. Code style linter is causing the CI to be red, but the specs do go through and that's nice. You might want to run `bundle exec rubocop` locally to see the errors on your machine, and if possible, configure your editor to show those errors inline to make the linter cause less anguish 😅 

The `AuthorizationStubs` are a bit difficult, I agree. I do however think that the solution with `.tap` is alright, and we can keep using that :relaxed:. It's too bad that we can't use `.to receive(operation).with(*args, **kwargs)` even if `kwargs` were empty.

I wonder if, after converting all authorizer methods to use kwargs, will we even need the "non-kwargs" call anyway? So we might be able to get rid of the `.tap`-dance (heh) after the refactor is complete? :tada:

nruth:
Thanks, I'll do that.

I'd like to mention a couple of other things, assuming semantic versioning:

- Renaming: some of the kw-args feel like they could be renamed/shorter. I've not done this initially because I don't feel I have any kind of ownership over the code and it'd make the changes harder to accept. Before 1.0 would however be the best time to consider this, because after that the API is more painful to change. I've not used this library much so I don't know how much changing this api would affect users, or if it's all internal code. I'm assuming it would affect some people who are writing their own authenticators, since the class name has "default" in it?

- Do all of the methods need keyword arugments? The ones with 1 argument and an obvious name will be easier to write with positional args. I personally prefer keyword arguments almost everywhere, and it might make the `*args, **kwargs` problem go away as you mentioned (depending on which other classes are tested with those helpers, and if they also get changed to kwargs). Making it consistent across all the methods in the class, and making it self-documenting, would probably be good things, but it is more typing or boilerplate – all very stylistic, so up to you really?

I'll continue with the simplest options for now, but as I mentioned it's probably better to get the API right (if that matters? if it's not frequently used maybe it does not) before a major release.

valscion:
Good question regarding the argument names.

I took a look at the current argument names, and only `related_records_with_context` in `replace_fields` and `create_resource` methods I'm unsure about. 

https://github.com/venuu/jsonapi-authorization/blob/da04d6cf5ca37c505e6503016141a4413e96c6b2/lib/jsonapi/authorization/default_pundit_authorizer.rb#L91

https://github.com/venuu/jsonapi-authorization/blob/da04d6cf5ca37c505e6503016141a4413e96c6b2/lib/jsonapi/authorization/default_pundit_authorizer.rb#L103

I think they'll be quite easy to rename after this PR to something else, so you don't have to worry about that here. That is, unless you get some amazing idea for a better name, then we can discuss it :smile:



> I'm assuming it would affect some people who are writing their own authenticators, since the class name has "default" in it?

Yeah, that has been the case. We've documented that people can use a different authorizer in the past here: https://github.com/venuu/jsonapi-authorization/blob/v0.8.2/README.md#configuration

I've also mentioned that subclassing `DefaultPunditAuthorizer` is a way to cater to different requirements, for example here https://github.com/venuu/jsonapi-authorization/issues/39#issuecomment-289378051. I think that @arcreative is using a custom authorizer class (seeing as he wrote this issue https://github.com/venuu/jsonapi-authorization/issues/77), subclassing `DefaultPunditAuthorizer`, so he might have insight into the argument names.

---



> Do all of the methods need keyword arugments? The ones with 1 argument and an obvious name will be easier to write with positional args.

I think it's for the best to use keyword arguments for all methods. It makes the authorizer class methods consistent, and also makes it clearer when the method is passed a model class compared to model instance. For example:

https://github.com/venuu/jsonapi-authorization/blob/da04d6cf5ca37c505e6503016141a4413e96c6b2/lib/jsonapi/authorization/default_pundit_authorizer.rb#L30

:point_up: the `#find` method gets a class, and:

https://github.com/venuu/jsonapi-authorization/blob/da04d6cf5ca37c505e6503016141a4413e96c6b2/lib/jsonapi/authorization/default_pundit_authorizer.rb#L39

:point_up: the `#show` gets a record / instance of a class.

---



> I'll continue with the simplest options for now, but as I mentioned it's probably better to get the API right (if that matters? if it's not frequently used maybe it does not) before a major release.

That sounds like a good approach to me! :+1: I don't think many people rely on the exact API of `DefaultPunditAuthorizer` (at least I hope so 😅) so changes done there shouldn't be an issue for most people.

spec/support/authorization_stubs.rb
```diff
@@ -3,7 +3,7 @@ module AuthorizationStubs
 
   def allow_operation(operation, *args, authorizer: instance_double(AUTHORIZER_CLASS), **kwargs)
     allow(authorizer).to receive(operation).tap {|x|
-      if kwargs.empty? then x.with(*args) else x.with(*args, **kwargs) end
+      kwargs.empty? ? x.with(*args) : x.with(*args, **kwargs)
```
nruth: I am a bit sad about this change, but I see the benefit of having the style rules enforced automatically & not spending time agonising over lots of custom rules (as I probably now will while adding this setup to my own project).

nruth:
I think this is finished, but would like your opinion on the unused parameters in `include_has_one_resource` and `include_has_many_resource`. Can these be deleted?

Otherwise, we either need to do this (what I've done currently, on 2 methods):

```
# rubocop:disable Lint/UnusedMethodArgument
def include_has_many_resource(source_record:, record_class:)
  ::Pundit.authorize(user, record_class, 'index?')
end
# rubocop:enable Lint/UnusedMethodArgument
```

or set `AllowUnusedKeywordArguments` to true in the rubocop config.

valscion:
Thanks! Looning forward to looking into this 😊. I'll get to this on Monday most likely

spec/support/authorization_stubs.rb
valscion: Yeah, code styles can differ between projects. I'm myself happy about this change, as I prefer ternaries over one line `if ... then ... else ... end` statements 😛

valscion:
I actually got to looking into this right now 😊.

This looks great to me! I think that it's the best approach to leave the rubocop disable comments in-place, as even if the default authorizer doesn't use those arguments, they can be useful for custom authorizer classes.

It's also possible to add only `rubocop:disable TheOffendingRule` to the end of a line to disable a rule for only that line. I don't really have a strong feeling one way or the other, so whichever approach suits you the most is fine by me :relaxed:.

valscion:
Is this OK for merging, or is there still something you'd like to do?

nruth:
I think the code is ready but I haven't checked if any of the documentation needs revising.

valscion:
Yeah might be useful to check if the generated rubydocs still look good. That can happen in another PR as well.

I don't remember off the top of my head that how these rubydocs were generated in the first place 😅

nruth:
The classname doesn't appear in the project files outside of the file comments and specs, so I assume there isn't other documentation for this. 

Upgrading user wise, the runtime documentation (*cough* error messages *cough*) should make it clear what to do. So I'm finished on this unless you'd like something else doing.

A nice thing for the changelog / version notes for v1 probably needs writing but I'm hoping you will do that?

valscion:
Yeah I can handle the changelog :relaxed:. Thanks a bunch for helping out! I'd like to add you to the contributors section in the README if that's OK to you?

nruth:
Great, thanks!

PR #96, opened 2018-06-30
Unlock/ignore dev ruby version & update CI ruby version

nruth:
This would update the dev and ci ruby version from an expired/EOL ruby 2.1 version to something newer. In fact, it lets people use whatever they want locally, and runs 2.3-2.5 on CI. I think this is nice because it would have meant as a new contributor I didn't need to wait to download/install ruby 2.1.2 and could have just used 2.4.3 or whichever I already had installed.

This reuses a commit from a previous pull request (nicer for the author). It looks ok to me, but I'm also going to open an alternative PR that only changes the ruby version.


valscion:
Niice :+1:. I just merged the other PR, and this improvement is super great as well. Thanks for doing this!

valscion:
Can you merge master here or rebase this on top of master? There seems to be conflicts and I'm on my mobile and fixing the conflict would be a bit difficult for me

nruth:
Yep will do.

nruth:
I've merged master, the conflicts were the travis file's ruby versions and the `.ruby-version` file, which both branches made changes to. Easy to fix, because this branch wins in both places.

Rails 5 isn't in the travis builds currently, but picking versions to test isn't obvious so I'm leaving that for another pr.

valscion:
Looks great to me!

For rails versions, I think testing the latest patch release of each minor release would be a good strategy.

So that would be:

* 5.0.7
* 5.1.6
* 5.2.0

If supporting 4.2 is causing trouble, we can also look into removing that from the build matrix.

Issue #101, opened 2018-07-19
Support for Pundit Policy Namespacing?


drewnichols:
Perhaps I have overlooked it but it seems that the way this gem is designed we are unable to take advantage of [Pundit's Policy Namespacing](https://github.com/varvet/pundit#policy-namespacing) feature. 

We have a motivation similar to https://github.com/venuu/jsonapi-authorization/issues/8 in which users can access our apis under different contexts, such as an admin or a user, and thus should receive different policies and different scopes. The namespacing seemed a good fit as we could pass the PolicyFinder an array containing the context and the object instead just the object and it will resolve policy classes as follows:

```
PolicyFinder.new([:admin, post]) # => Admin::PostPolicy 
PolicyFinder.new([:user, post])  # => User::PostPolicy
```

Was this intentional? Am I missing something?

Subtletree:
I would also be hugely in favor of this.

Namespaced policies would fit perfectly with namespaced resources/controllers imo.

@valscion Would this just require a processor change?

valscion:
Hi, thanks for opening an issue! Sorry for taking so long to reply — I've been on two weeks vacation and just came back to work today.

It seems like `jsonapi-authorization` hasn't been designed for use with policy namespacing the way Pundit does it. Policy namespacing seems to have been documented for v2.x of Pundit https://github.com/varvet/pundit/commit/ef5cc85f9d254492f7215425d4032d82596329bb and this library was written with Pundit v1, so this feature hasn't been considered when this gem was written.

I see @Subtletree has created a proof of concept for automatic namespacing based on resource class names in #102. I think it would be better if we didn't look at the resource class names and then use the namespace from that.

It would be nice to see some design around an explicit policy class definition, for example by using a special resource class method as originally proposed in #8 if that would fit your use case.

Subtletree:
Hope you had a good break @valscion 😄 

I think that's a good idea to step back to a design stage and look at actual use cases for namespaced resources/policies.

If there are use cases for specifying a policy override per resource then that sounds like the way to go.

If it ends up being that namespaced policies are always wanted for namespaced resources then I think automatic namespacing should be the way to go. This is how jsonapi-resources handles matching controllers to resources. Namespaced controllers are automatically matched to namespaced resources [via the controller class name](https://github.com/cerebris/jsonapi-resources/blob/87db4483d387b68e02fef8a16f39af8adb6134f9/lib/jsonapi/acts_as_resource_controller.rb#L137-L139).

[...]

PR #102, opened 2018-07-30
Use namespaced policies

Subtletree:
Proof of concept for namespaced policies #101 #8 

If this looks like something that will be merged, I'll finish up with:

- Finish tests (maybe slim down and just test namespacing rather than duplicating other tests and adding namespace)
- Add something to the readme.
- Update method comments

cc @drewnichols

### DIFF
--- a/jsonapi-authorization.gemspec
+++ b/jsonapi-authorization.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency "jsonapi-resources", "~> 0.9"
-  spec.add_dependency "pundit", "~> 1.0"
+  spec.add_dependency "pundit", "~> 2.0"
 
   spec.add_development_dependency "bundler", "~> 1.11"
   spec.add_development_dependency "rake", "~> 12.0"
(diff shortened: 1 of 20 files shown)


Subtletree:
Updated minimum pundit version to 2.0 https://github.com/varvet/pundit/pull/529

jpalumickas:
Is it possible to make this as optional? (Maybe enable/disable in initializers)
For example, we have resources under `Api::V1` namespace (`Api::V1::UserResource`), but our policies are without namespace (`UserPolicy`), so I think this PR will break, or no?

valscion:
@jpalumickas see the discussion in #101 — this PR will definitely not be merged as it currently stands :relaxed:

jpalumickas:
@valscion oh! thanks for info 😉 Is it possible easily to upgrade this gem to support Pundit v2? I don't think there was lots of breaking changes there?

valscion:
> Is it possible easily to upgrade this gem to support Pundit v2? I don't think there was lots of breaking changes there?

Might be! I opened an issue to discuss this further: https://github.com/venuu/jsonapi-authorization/issues/103

valscion:
Hi! I'll close this PR for now as this isn't the approach we would like to go with. Feel free to continue the discussion in #103.

Thanks for creating a proof of concept of this approach :heart:

PR #107, opened 2019-01-07
Fix CI caused by Bundler v2

valscion:
Seems like CI is failing because Travis is using bundler v2:
https://travis-ci.org/venuu/jsonapi-authorization/jobs/415322974#L468


valscion:
I closed the PR before the other Travis build ran, so the CI shows a failure because of that. The CI run triggered by a push however completes successfully: https://travis-ci.org/venuu/jsonapi-authorization/builds/476260382

PR #109, opened 2019-01-07
Remove support for Rails 4.2

valscion:
We no longer use Rails 4.2 at work, so this gem can't support that version either


valscion:
Oh right, can't do that, as `jsonapi-resources` seemingly still supports Rails 4.2

https://github.com/cerebris/jsonapi-resources/blob/v0.9.4/README.md

> JR is designed to work with Rails 4.2+, and provides custom routes, controllers, and serializers. JR's resources may be backed by ActiveRecord models or by custom objects.

And they test for 4.2 as well:

https://github.com/cerebris/jsonapi-resources/blob/v0.9.4/.travis.yml#L3-L6

Oh well.

Issue #118, opened 2019-03-12
Updating related resources does not check against the related records


brianswko:
When updating resources e.g. `PATCH /articles/:id/relationships/comments` or `PATCH /articles/:id` with related resources that the user does _not_ have access to, I would expect either a 403 or 404 response instead of a 2xx.

https://github.com/venuu/jsonapi-authorization/blob/c7b69079be557848083e7aab77627f95a266f22e/spec/requests/relationship_operations_spec.rb#L179-L182 is marked as a pending spec with a not found response, but it currently returns a 204 ~and updates the article with comments that the user does not have access to.~ edit: this was a problem with the spec, not the code itself

https://github.com/venuu/jsonapi-authorization/blob/c7b69079be557848083e7aab77627f95a266f22e/spec/requests/tricky_operations_spec.rb#L212 explicitly checks that if a user sends a PATCH to an article with comments that the user does not have access to, that a successful response is returned (and though the spec doesn't test it, the article now includes those comments that the user doesn't have access to.

I've confirmed that even when writing a custom `ArticlePolicy#replace_comments?` method, the comments that are passed in are already limited to those that are in scope by https://github.com/venuu/jsonapi-authorization/blob/c7b69079be557848083e7aab77627f95a266f22e/lib/jsonapi/authorization/authorizing_processor.rb#L335 so the method never has a chance to raise an error or return false.

I believe this is a bug, and that it started with https://github.com/venuu/jsonapi-authorization/commit/ad9aad277d6d262b2b1a2268d18c056362d75294 when `related_models_with_context` was added and used `find_by_keys` instead of `_model_class.where(...)`. I'd change it back in a PR but am uncertain what other ramifications that may have.

Thoughts?

valscion:
Thank you for the detailed report. This does seem like a bug.

Checking the spec you highlighted more closely:

https://github.com/venuu/jsonapi-authorization/blob/c7b69079be557848083e7aab77627f95a266f22e/spec/requests/relationship_operations_spec.rb#L173-L183

It should assert that `replace_to_many_relationship` was called with all the new comments and not an empty array. Is that not the case? I'm no longer certain if the keywords must all match for `allow_operation`, but it seems like they must or else the authorization stubbing would not work?

https://github.com/venuu/jsonapi-authorization/blob/c7b69079be557848083e7aab77627f95a266f22e/spec/support/authorization_stubs.rb#L4-L9

brianswko:
[...]

PR #119, opened 2019-03-21
Check for authorization against related records

brianswko:
Fixes https://github.com/venuu/jsonapi-authorization/issues/118

This PR fixes a bug where related resources are not checked for authorization if they are not in the user's scope.

Also fixes inconsistencies between the behavior of has-one, has-many, and polymorphic relationships to always return a 404 instead of sometimes returning a 403 when the user should not be able to see that a given record exists.

### DIFF
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -209,15 +209,10 @@ module JSONAPI
 
         relationship_type = params[:relationship_type].to_sym
 
-        related_resources = @resource_klass
-          ._relationship(relationship_type)
-          .resource_klass
-          .find_by_keys(
-            params[:associated_keys],
-            context: context
-          )
-
-        related_records = related_resources.map(&:_model)
+        related_resource_class = @resource_klass._relationship(relationship_type).resource_klass
+        related_records = related_resource_class._model_class.where(
+          related_resource_class._primary_key => params[:associated_keys]
+        )
 
         authorizer.remove_to_many_relationship(
           source_record: source_record,
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -329,13 +329,12 @@ module JSONAPI
                 nil
               when Hash # polymorphic relationship
                 resource_class = @resource_klass.resource_for(assoc_value[:type].to_s)
-                resource_class.find_by_key(assoc_value[:id], context: context)._model
-              when Array
-                resource_class = resource_class_for_relationship(assoc_name)
-                resource_class.find_by_keys(assoc_value, context: context).map(&:_model)
+                primary_key = resource_class._primary_key
+                resource_class._model_class.where(primary_key => assoc_value[:id])
               else
                 resource_class = resource_class_for_relationship(assoc_name)
-                resource_class.find_by_key(assoc_value, context: context)._model
+                primary_key = resource_class._primary_key
+                resource_class._model_class.where(primary_key => assoc_value)
               end
 
             {
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -298,25 +298,6 @@ module JSONAPI
         resource_class_for_relationship(assoc_name)._model_class
       end
 
-      def related_models
-        data = params[:data]
-        return [] if data.nil?
-
-        [:to_one, :to_many].flat_map do |rel_type|
-          data[rel_type].flat_map do |assoc_name, assoc_value|
-            case assoc_value
-            when Hash # polymorphic relationship
-              resource_class = @resource_klass.resource_for(assoc_value[:type].to_s)
-              resource_class.find_by_key(assoc_value[:id], context: context)._model
-            else
-              resource_class = resource_class_for_relationship(assoc_name)
-              primary_key = resource_class._primary_key
-              resource_class._model_class.where(primary_key => assoc_value)
-            end
-          end
-        end
-      end
-
       def related_models_with_context
         data = params[:data]
         return { relationship: nil, relation_name: nil, records: nil } if data.nil?
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -209,10 +209,18 @@ module JSONAPI
 
         relationship_type = params[:relationship_type].to_sym
 
-        related_resource_class = @resource_klass._relationship(relationship_type).resource_klass
-        related_records = related_resource_class._model_class.where(
-          related_resource_class._primary_key => params[:associated_keys]
-        )
+        related_resources = @resource_klass
+          ._relationship(relationship_type)
+          .resource_klass
+          .find_by_keys(
+            params[:associated_keys],
+            context: context
+          )
+        related_records = related_resources.map(&:_model)
+
+        if related_records.count != params[:associated_keys].uniq.count
+          fail JSONAPI::Exceptions::RecordNotFound.new(params[:associated_keys])
+        end
 
         authorizer.remove_to_many_relationship(
           source_record: source_record,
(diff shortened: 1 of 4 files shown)


valscion:
Sorry about Travis CI failing ...again... we've fixed the errors caused by bundler upgrade several times now but it seems they keep on failing.

valscion:
I fixed the CI in https://github.com/venuu/jsonapi-authorization/pull/120 and merged `master` here. Let's see what the CI says now.

valscion:
CI seems to be happy. And so am I :relaxed:

This seems like it's ready to go, right? How would we describe this change, similarly to how we've done change docs in the past? https://github.com/venuu/jsonapi-authorization/releases

@all-contributors please add @brianswko for bug, test and code

valscion:
@all-contributors please add @brianswko for bug, test and code

Let me know if you'd rather not be listed in the README. I'll wait for your confirmaton before merging the README.md update by the bot.

spec/requests/tricky_operations_spec.rb
```diff
@@ -57,6 +57,12 @@
       before { disallow_operation('create_resource', source_class: Comment, related_records_with_context: related_records_with_context) }
 
       it { is_expected.to be_forbidden }
+
+      context 'which is out of scope' do
+        let(:policy_scope) { Article.none }
+
+        it { is_expected.to be_forbidden }
```
brianswko: @valscion I just realized while testing that instead of being forbidden this should probably be a not found response. I'm going to rework the processor a bit.

valscion: Super :relaxed:. Take your time

brianswko:
Proposed description:

Fixes PATCH and POST requests to check if the user has the correct permissions for every given object in a has-many relationship

For example:
If a user does not have access to (meaning the pundit scope does not include) the author with ID 2
i.e. `AuthorPolicy::Scope.new(user, Author).resolve.include?(Author.find(2)) # => false`
And the following request is called:
```
PATCH /books/1

"data": {
  "type": "books",
  "id": "1",
  "attributes": {...},
  "relationships": {
    "authors": {
      "data": [
        { "type": "authors", "id": "1" },
        { "type": "authors", "id": "2" }
      ]
    }
  }
}
```
Previously: Would return a 20x and update the book to include author 2
Now: Will return a 404 and not update the book since the user is unable to find author 2

In some scenarios, this will cause a 404 to be returned where a 403 used to be returned.

brianswko:
@valscion I've simplified the change in the processor, but ended up having to make a decent amount of changes to the specs. Hoping those all make sense to you.

I'm conflicted as to whether this should be a major or minor version change, as there will almost certainly be a not-insignificant number of projects that will start seeing specs fail as bugs in specs are exposed and formerly incorrect 403s start returning 404s. Will leave that decision in your hands.

spec/requests/included_resources_spec.rb
```diff
@@ -7,20 +7,20 @@
...
-    )
+    allow_any_instance_of(ArticlePolicy::Scope).to receive(:resolve) do |policy_scope|
+      article_policy_scope.merge(policy_scope.scope.all)
+    end
+    allow_any_instance_of(CommentPolicy::Scope).to receive(:resolve) do |policy_scope|
+      comments_policy_scope.merge(policy_scope.scope.all)
+    end
+    allow_any_instance_of(UserPolicy::Scope).to receive(:resolve) do |policy_scope|
+      user_policy_scope.merge(policy_scope.scope.all)
+    end
```
valscion: I'm having a bit of a hard time figuring out what these policy scope changes do. It seems like it's overriding the policy scope stubbing to always return all results?

Maybe there could be some sort of a documentation comment here to tell why this is necessary.

spec/requests/included_resources_spec.rb
```diff
@@ -58,11 +56,11 @@
 
         it { is_expected.to be_successful }
 
-        let(:comments_policy_scope) { Comment.limit(1) }
-
-        it 'includes only comments allowed by policy scope' do
-          expect(json_included.length).to eq(1)
-          expect(json_included.first["id"]).to eq(comments_policy_scope.first.id.to_s)
+        it 'includes only comments allowed by policy scope and associated with the article' do
+          expect(json_included.length).to eq(article.comments.count)
```
valscion: Hmm so does this now test that there are more records than were allowed? I.e. do we lose test coverage on the pundit policy scope being applied?

lib/jsonapi/authorization/authorizing_processor.rb
```diff
@@ -219,6 +219,10 @@ def authorize_remove_to_many_relationships
 
         related_records = related_resources.map(&:_model)
 
+        if related_records.count != params[:associated_keys].uniq.count
```
valscion: Is there any way `.count` would accidentally trigger an SQL `COUNT` query even if a cached version would exist in these changes? I.e. should we use `.size` instead of `.count` here? Does it matter?

valscion:
> Fixes PATCH and POST requests to check if the user has the correct permissions for every given object in a has-many relationship

Nice. This is good.



> I'm conflicted as to whether this should be a major or minor version change, as there will almost certainly be a not-insignificant number of projects that will start seeing specs fail as bugs in specs are exposed and formerly incorrect 403s start returning 404s. Will leave that decision in your hands.

Given we only recently upgraded to 2.0.0 to fix an authorization bug, and only less than 1000 downloads as of this date have been done against [v2.0.0](https://rubygems.org/gems/jsonapi-authorization/versions/2.0.0), we could do this change and just bump to 3.0.0.

I'd rather signal a potentially backwards incompatible change with more major versions than to squint my eyes and think this could be a backwards-compatible change.

So once this is ready to be merged, let's punt this to 3.0.0.

lib/jsonapi/authorization/authorizing_processor.rb
brianswko: Since `related_records` is created by calling `map` on a collection of Resources, it will always be an array and not trigger any SQL queries. On the other hand, I forgot that `Array#size` is more performant than `count` so I'll change that here.

spec/requests/included_resources_spec.rb
brianswko: I wanted to get rid of these entirely, but that required a much larger overhaul of the tests than I wanted to take on.

As previously written, this spec never checked which `has_many` records were returned except in situations where the spec stubbed the `comments_policy_scope` to be `Comment.limit(1)`, and it masked a fundamental problem with the mock when I added additional scenarios to test.

https://github.com/venuu/jsonapi-authorization/blob/05785ff8c3d841eae2a959c59f80b9ad408c5214/lib/jsonapi/authorization/pundit_scoped_resource.rb#L15-L28
The value of `policy_scope!` is supposed to be the records that the user has access to, further restricted by the specific `record_or_records` in the association. A simple `and_return` like what was here before would return all records in the user's scope, even if it had nothing to do with the `@model` under test. For example, a user who has access to all articles and looked up the articles written by a single user would get a list of all articles instead of just the articles written by that user.

`merge` acts like a sql `AND`, so this change allows the stubbed scope to return something more similar to how it should actually behave outside of tests.

Will add a comment for this.

spec/requests/included_resources_spec.rb
brianswko: I moved the specs specifically testing for limited scopes to https://github.com/venuu/jsonapi-authorization/pull/119/files#diff-f3a557fdd906ce389d7b5d190c721b99R220. I had to move them out of this shared spec since the behavior is different for GET (which returns the relationships but limited by the scope) vs POST (which should return a 404).

valscion:
This looks like it's ready to go for me! What do you think? Is it ready?

I'll update the README to clarify what's the intent on version numbering here: https://github.com/venuu/jsonapi-authorization/pull/123. I think it's likely to be helpful given how this change will be shipped as v3.0.0 once we merge it.

I'll copy the text of https://github.com/venuu/jsonapi-authorization/pull/119#issuecomment-476550980 for the release notes. Let me know if there's any changes you'd like there to be.

spec/requests/included_resources_spec.rb
valscion: Perfect, thanks for clarifying!

spec/requests/included_resources_spec.rb
valscion: Nice, thanks!

brianswko:
Yes it's ready to go, and that comment works for the changelog.

valscion:
Thank you! This has now been releases as [v3.0.0](https://github.com/venuu/jsonapi-authorization/releases/tag/v3.0.0) :tada:

PR #124, opened 2019-03-29
Fix a bug with has-one relationship authorizers

brianswko:
Fixes a bug introduced in https://github.com/venuu/jsonapi-authorization/pull/119

Any has-one relationship authorizers e.g. `ArticlePolicy#create_with_author?(author)` are currently breaking since the default authorizer is calling the method with `[author]` instead of `author`.

I was trying to be too clever in that PR simplifying code and didn't realize that there may be code elsewhere that relied on has-one relationships sending related records back as an actual record instead of potentially a single-object array.

I'd like to add a spec to cover this case since there are no (non-stub setup reliant) specs covering this behavior, but am unsure how to structure it. I believe this can go out on its own now, or am happy to take some feedback as to where to add a test for this.

While there was a major version bump to signify breaking changes, this was not an intended breaking change.

### DIFF
--- a/lib/jsonapi/authorization/authorizing_processor.rb
+++ b/lib/jsonapi/authorization/authorizing_processor.rb
@@ -315,7 +315,7 @@ module JSONAPI
               when Hash # polymorphic relationship
                 resource_class = @resource_klass.resource_for(assoc_value[:type].to_s)
                 resource_class.find_by_key(assoc_value[:id], context: context)._model
-              else
+              when Array
                 resource_class = resource_class_for_relationship(assoc_name)
                 resources = resource_class.find_by_keys(assoc_value, context: context)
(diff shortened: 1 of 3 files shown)


valscion:
Thanks!

One way to go about without stubbing things could be to create some sort of minitest-like spec where all the relevant classes would be defined in the test file itself. Then those tests could go through all the layers without relying on any stubbing mechanisms.

Even having one such test would be a huge boon to the development ability.

I'd be interested to see how it would look like with minitest. RSpec would be also be OK. If only we could have all the relevant classes under test defined in the same file, that'd be nice — once that file gets unwieldy, we could figure out how to split it.

valscion:
I released this as v3.0.1 — and would be amazing to have tests that wouldn't rely on stubbing 😅

PR #128, opened 2019-10-02
Remove implicit block expectation syntax

brianswko:
While attempting to run specs to investigate compatibility with JR `v0.10.0`, I ran into several failures with the newly released `rspec-expectations` `v3.8.5`

https://github.com/rspec/rspec-expectations/pull/1125 removed the ability to use implicit block expectations


valscion:
Huh, that's a strange change to be doing in a patch release indeed.

I'll take a look when I get to a machine that doesn't freeze when looking at the diff like my mobile does 😅

PR #129, opened 2019-10-03
Run CI against Rails 6

brianswko:
Loosens the `sqlite3` dev dependency to allow tests to run on any supported version of rails.

Adds appraisals and gemfiles for rails 6 to run in CI

Updates the ruby version in CI to 2.5
Resolves #132

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ branches:
     - master
 rvm:
   - 2.3
+  - 2.5
 gemfile:
   - gemfiles/rails_4_2_pundit_1.gemfile
   - gemfiles/rails_5_0_pundit_1.gemfile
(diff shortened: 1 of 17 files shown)


.travis.yml
```diff
@@ -4,15 +4,28 @@ branches:
...
   - gemfiles/rails_5_1_pundit_1.gemfile
   - gemfiles/rails_5_2_pundit_1.gemfile
+  - gemfiles/rails_6_0_pundit_1.gemfile
   - gemfiles/rails_4_2_pundit_2.gemfile
   - gemfiles/rails_5_0_pundit_2.gemfile
   - gemfiles/rails_5_1_pundit_2.gemfile
   - gemfiles/rails_5_2_pundit_2.gemfile
+  - gemfiles/rails_6_0_pundit_2.gemfile
+matrix:
+  exclude:
```
brianswko: This is the minimum number of exclusions since rails 4.2 doesn't like ruby 2.5 and rails 6 doesn't support ruby 2.3. Totally open to adding more exclusions so we're not running 16 builds

valscion: Is it possible to only run Ruby 2.5 for Rails 6 runs, and keep 2.3 for other Rails versions?

PR #133, opened 2020-01-12
Update build badge to point to travis-ci.com

valscion:
This repository has been migrated from the old https://travis-ci.org to https://travis-ci.com and is now running with the Travis GitHub application.

Let's see if this brings better status checks to pull requests :relaxed:


valscion:
Yup, works :relaxed:. New builds will have to be triggered on open pull requests so that the status checks do not refer to the old travis-ci.org status checks that are no longer happening.

PR #138, opened 2021-01-15
Github Actions Take 2

cloke:
This a second PR to test GitHub actions from a branch other than master.


cloke:
@venuu I invited you directly to my fork. You can see the actions run there (but fail). Even after removing the travis file it still seems to run that over actions. Maybe there is a setting I am not seeing.

valscion:
Hmm. I can see that the GitHub action does run in #139.

The bundler and rake dependencies should not be updated as otherwise the tests won't run with the gemfiles we have.

I invited you to this repository if you'd like to try out getting the `github-actions` branch working directly on this repository to get #139 working. That PR at least runs the GitHub actions.

PR #139, opened 2021-01-18
GitHub actions test run

valscion:
Commit from https://github.com/venuu/jsonapi-authorization/pull/138

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -29,6 +29,7 @@ jobs:
     # change this to (see https://github.com/ruby/setup-ruby#versioning):
       uses: ruby/setup-ruby@v1
       with:
+        bundler: 1.11
         ruby-version: ${{ matrix.ruby-version }}
         bundler-cache: true # runs 'bundle install' and caches installed gems automatically
     - name: Run tests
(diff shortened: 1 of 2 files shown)


cloke:
It looks like the main issue will be working through bundle versions versus ruby versus rails versions. For example, if I run the test against the 4.2 rails version using bundle 1.11 there is a Nokogiri error. If I run bundle 1.11 against the rails 6 version then it fails in other ways. As I am not super familiar with how the appraisals gem works I will need to come back to this later.

valscion:
Yeah looks like the CI will fail with Rails 4.2. We can drop support for that in the next release so we don't have to get CI working with Rails 4.2.

Looks like the existing Travis steps do run after all, but they fail with Rails 4.2: https://github.com/venuu/jsonapi-authorization/runs/1719931321

Skip_bumping_bundler_and_rake_dependency_·_venuu_jsonapi-authorization_e189c5c

cloke:
Some additional findings...

Rails 6 tests will pass with Ruby 3 and Ruby 2.7.
Rails 5 tests fail with Ruby 3 which is probably expected. I'm still learning how Git Actions work with versions. Ideally we'd specify tests to use ruby "<3" on rails "<6".

valscion:
We don't need to run with Ruby 3 yet, let's get 2.x Ruby versions running first ☺️ we can then look at adding test suite with Ruby 3.0 in the future once we get the GitHub actions working ☺️

PR #140, opened 2021-02-08
Github Actions To Replace Travis

cloke:
Third attempt to get all actions to work with current supported ruby / rails / bundle versions.


valscion:
Cool! Let me know when this is ready for review ☺️

encounter:
This is ready for review. We have all the existing tests passing, plus Ruby 3.0 + Rails 6.0.

valscion (approved):
Amazing! This looks good to me! :+1:

It's OK to keep the Travis CI still running, we can get rid of that later if for some reason it happens to break ☺️.

PR #152, opened 2024-07-24
Fix returning a not_found for has_one relation

Skipants:
When an included resource does not exist in the Policy Scope for that resource, it's impossible to update the base resource even if we try to circumvent it with `replace_resource?` on the base resource's policy.

eg. Consider we have the following policies:

class FooPolicy < ApplicationPolicy
  def replace_bar(_bar)?
    true
  end
end

class BarPolicy < ApplicationPolicy
  class Scope < Scope
    def resolve
      scope.none
    end
  end
end

When we try to update a Foo resource with a Bar resource, the `related_models_with_context` method on the `AuthorizingProcessor` will return a not_found error bubbled up from jsonapi-resources because the Bar resource does not exist in the policy scope.

The expected behavior is update Foo with a Bar without caring about the Bar resource's policy scope.


Skipants:
Oops -- sorry. Meant to push this to our repo, not this one.

